PR #4, opened 2023-04-26
Fix: add various setup improvements

TheDZhon:
- Add MIT license
- Enrich the `pyproject.toml` contents
- Use `ETHERSCAN_TOKEN` instead of `ETHERSCAN_API_TOKEN` for compatibility with web3 frameworks such as brownie
- Provide setup instructions and list dependencies
- Exit early if a single contract requested without its name

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,8 +9,8 @@ Diff your Ethereum smart contracts code from GitHub against Etherscan verified s
 
 ## Prerequisites
 This project was developed using these dependencies with their exact versions listed below:
-- Python ^3.10
-- Poetry ^1.4
+- Python 3.10
+- Poetry 1.4
 
 Other versions may work as well but were not tested at all.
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -1,12 +1,50 @@
...
+pip install --user poetry~=1.4
+```
+
+alternatively, you could proceed with `pipx`:
+
+```shell
+pipx install poetry~=1.4
+```
+
+2. Install Python dependencies
```
mymphe, line 33: Not sure why but `poetry install` => `poetry shell` sometimes fails for me in some poetry projects. On the other hand, `poetry shell` first => `poetry install` works every time

mymphe (approved):

folkyatina (approved):
basically LGTM

README.md
TheDZhon: Np, changed the order :+1:

PR #5, opened 2023-04-27
feat: resolve deps

mymphe:


utils/github.py
```diff
@@ -25,3 +26,32 @@ def get_file_from_github(github_api_token, dependency_repo, filepath):
...
+
+
+def construct_filepath(filepath):
+    # return the same thing if the file is in the local repo
+    if is_local_file(filepath):
+        return filepath
+
+    # extract "contracts..." substring (preceded by "/") from filepath
+    # e.g. "@openzeppelin/contracts-v4.4/utils/structs/EnumerableSet.sol" => "contracts-v4.4/utils/structs/EnumerableSet.sol"
+    regex = re.compile("(?<=\/)contracts.*")
```
TheDZhon, line 38: Can we avoid regex and just define stupid remappings in the config?

TheDZhon (commented):
:thinking:

utils/github.py
TheDZhon: I was thinking about it in this way:
1) our remapping is always a prefix, not an arbitrary substring in the middle
2) we can sort remappings based on prefix length and apply them starting from the longest ones, returning early if found
3) maybe it would be great to require every remapping ends with closing '/'

Doing all of this (or at least 1+2) would allow us to avoid regular expressions at all

config.sample.json
```diff
@@ -2,14 +2,16 @@
...
     "network": "mainnet",
     "github_repo": "https://github.com/lidofinance/lido-dao/tree/master",
     "dependencies": {
-        "@aragon": "https://github.com/aragon/aragonOS/tree/f3ae59b00f73984e562df00129c925339cd069ff",
+        "@aragon/os": "https://github.com/aragon/aragonOS/tree/f3ae59b00f73984e562df00129c925339cd069ff",
         "openzeppelin-solidity": "https://github.com/OpenZeppelin/openzeppelin-contracts/tree/06e265b38d3e9daeaa7b33f9035c700d6bc0c6a0",
-        "solidity-bytes-utils": "https://github.com/GNSPS/solidity-bytes-utils/tree/9776282d181839fbb4b18f2cf218e316d6df871c"
+        "solidity-bytes-utils": "https://github.com/GNSPS/solidity-bytes-utils/tree/9776282d181839fbb4b18f2cf218e316d6df871c",
+        "@openzeppelin/contracts": "https://github.com/OpenZeppelin/openzeppelin-contracts/tree/v3.4.0",
+        "@openzeppelin/contracts-v4.4": "https://github.com/OpenZeppelin/openzeppelin-contracts/tree/v4.4.1"
```
TheDZhon, line 15: maybe use commits and not tags, IDK

main.py
```diff
@@ -80,6 +78,9 @@ def run_diff(config, name, address, etherscan_api_token, github_api_token):
 
             diffs = difflib.unified_diff(github_lines, etherscan_lines)
             diffs_count = len(list(diffs))
+        else:
```
TheDZhon, line 81: can move this one upper (just after we got `repo`) to exit early

utils/github.py
```diff
@@ -25,3 +26,34 @@ def get_file_from_github(github_api_token, dependency_repo, filepath):
         logger.error("No file content")
 
     return base64.b64decode(file_content).decode()
+
+
+def construct_filepath(filepath, dep_name):
```
TheDZhon, line 31: can check that `dep_name` is `None`

TheDZhon (approved):
LGTM :+1: 

Please take a look at the suggestions.

TheDZhon (approved):
:+1:

PR #9, opened 2023-05-02
Stop rewriting diffs for contracts with matching names

folkyatina:


main.py
```diff
@@ -1,4 +1,5 @@
 import difflib
+from os import name
```
mymphe, line 2: I think this thing got mistakenly auto-imported

PR #17, opened 2023-10-04
Replace network parameter with indexer host in config

kovalgek:
### WHAT
Setting network name isn't enough for some indexers. For instance base has **.org** name instead of **.io**.
 
#### HOW
Use indexer host as parameter to make **diffyscan** more flexible.

### DIFF
--- a/main.py
+++ b/main.py
@@ -13,7 +13,7 @@ from utils.logger import logger
 def run_diff(config, name, address, etherscan_api_token, github_api_token):
     logger.divider()
     logger.okay("Contract", address)
-    logger.okay("Indexer Host", config["indexer_host"])
+    logger.okay("Etherscan Hostname", config["etherscan_hostname"])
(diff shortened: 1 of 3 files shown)


main.py
```diff
@@ -13,15 +13,15 @@
 def run_diff(config, name, address, etherscan_api_token, github_api_token):
     logger.divider()
     logger.okay("Contract", address)
-    logger.okay("Network", config["network"])
+    logger.okay("Indexer Host", config["indexer_host"])
```
TheDZhon, line 16: `Etherscan Hostname`

mymphe (approved):
LGTM!

The proposed change seems to make Diffyscan more flexible, however, this makes the network implicit which could potentially lead to some misconfiguration bugs.

TheDZhon:
@mymphe @kovalgek maybe we should have two parameters?
`network` and `etherscan_host`?

PR #25, opened 2023-10-24
feat: add recursive dependency search for imported contracts

rkolpakov:
Added possibility to verify contracts if they have dependencies. Used a recursive search over the repo on github to cover complex contracts e.g. EasyTrack.


TheDZhon:
Thank you for proposing this change :pray: 

Let me shed a bit of light here :thinking: 
So the issue arose with the `brownie` verification tooling (i.e., they rewrite the imports on the sources submission)

See the attached screens:
photo_2023-10-24_19-52-46
photo_2023-10-24_19-52-45

I suggest NOT enabling recursive ('fuzzy') search by default but instead utilizing it only if the enforcement flag is provided within the config. Also, the flag might be omitted to preserve backward compatibility, falling back to the previous (strict search) behavior.

wdyt, @rkolpakov and @mymphe?

arwer13:
I'd prefer it enabled by flag (NOT default) as well. Although I cannot come up with an example when enabling this by default would lead to a bad outcome, I'd like to see more strictness in such a bedrock tool.

mymphe:
Agree with @arwer13. I think this feature is only useful for certain scenarios and should be explicitly enabled. Overall, a tool such as Diffyscan should be as dumb as possible, imo. Making it too smart might lead to critical false positives.

utils/explorer.py
```diff
@@ -21,6 +21,8 @@ def _get_contract_from_etherscan(token, etherscan_hostname, contract):
         logger.error("Not a contract or source code is not verified", contract)
         sys.exit(1)
 
+    print(123, data["SourceCode"])
```
arwer13, line 24: debug?

utils/arguments.py
```diff
@@ -0,0 +1,8 @@
+import argparse
+
+parser = argparse.ArgumentParser()
+parser.add_argument(
+    "--support-brownie",
+    help="Support recursive retrieving for contracts based on brownie framework.",
```
arwer13, line 6: Please provide more context in the help description regarding the use case. Could take some info from https://github.com/lidofinance/diffyscan/pull/25#issuecomment-1777658907

utils/github.py
```diff
@@ -24,6 +24,67 @@ def get_file_from_github(github_api_token, dependency_repo, path_to_file, dep_na
...
+    
+    direct_file_content = _get_direct_file(github_api_token, user_slash_repo, dependency_repo['relative_root'], path_to_file, dependency_repo['commit'])
+    if direct_file_content:
+        return direct_file_content
+    
+    return _recursive_search(github_api_token, user_slash_repo, dependency_repo['relative_root'], path_to_file, dependency_repo['commit'])
+
+
+def _get_direct_file(github_api_token, user_slash_repo, relative_root, path_to_file, commit):
+    github_api_url = f"https://api.github.com/repos/{user_slash_repo}/contents/{relative_root}/{path_to_file}?ref={commit}"
```
arwer13, line 39: maybe separate into a separate function these 2 lines of code to fetch from github api which common for `_get_direct_file` and `_recursive_searach`

arwer13 (changes requested):

README.md
```diff
@@ -88,4 +88,12 @@ Start the script
 python3 main.py
 ```
 
+> Note: Brownie verification tooling might rewrite the imports in the source submission. It transforms relative paths to imported contracts into flat paths ('./folder/contract.sol' -> 'contract.sol'), which makes Diffyscan unable to find a contract for verification.
+
+For brownie based projects start script with flag:
```
arwer13, line 93: Maybe "for contracts whose sources were verified by brownie tooling"...? Because contracts from brownie based project could be verified in any other way as well

utils/arguments.py
arwer13, line 6: Let's improve this as well. Maybe:
 ```
 help="Support recursive retrieving for contracts. It may be useful for contracts whose sources have been verified by the brownie tooling, which automatically replaces relative paths to contracts in imports with plain contract names.",
 ```

TheDZhon (approved):

PR #31, opened 2023-12-05
feat: + cmd args: path to config(-s) and -y to skip user confirmation

arwer13:
Add abilities to:
- set `--yes or -y` command argument to disable promtping the user to continue with the next contract verification
- optional positional argument: path to config or path to directory with configs
- old env variables inputs are intact


main.py
```diff
@@ -112,45 +125,60 @@ def run_diff(config, name, address, explorer_api_token, github_api_token):
...
-
-    logger.info("Loading config...")
-    config = load_config()
+def process_config(path: str):
+    logger.info(f"Loading config {path}...")
+    config = load_config(path)
 
-    if contract_address is not None:
-        if contract_name is None:
+    if CONTRACT_ADDRESS is not None:
```
TheDZhon, line 132: Let's get rid of this at all (CONTRACT_ADDRESS, CONTRACT_NAME from env)

main.py
```diff
@@ -1,15 +1,26 @@
...
 from utils.common import load_config, load_env
-from utils.constants import DIFFS_DIR, START_TIME
+from utils.constants import DIFFS_DIR, START_TIME, DEFAULT_CONFIG_PATH
 from utils.explorer import get_contract_from_explorer
 from utils.github import get_file_from_github, resolve_dep
 from utils.helpers import create_dirs
 from utils.logger import logger
 
 
+EXPLORER_API_TOKEN = load_env("ETHERSCAN_TOKEN", masked=True)
```
TheDZhon, line 15: might want to have multiple tokens (Optimism would require their own, Linea allows empty token for now, but they maintain a separate database and can change it in the future)

TheDZhon (changes requested):
LGTM, and huge thanks!

though I would polish this a bit :pray:

main.py
arwer13: with pleasure

main.py
arwer13: Added to config field which specified token env variable name.
Btw, the token happens not to be required for Etherscan as well (at least for our number of requests)

PR #32, opened 2024-01-16
fix: typo in mantle mainnet config sample

arwer13:


arwer13:
closed because the fix is going to be absorbed by another PR

PR #33, opened 2024-01-17
Feat/l2 target lidofinance

arwer13:
Update linea, mantle, zksync sample configs: make l2 repo references target repos in lidofinance

Also got sure the commits in configs correspond to the commits in the audits.

NB linea testnet not updated because it is cumbersome on Goerli is going to die

Plus two small fixes


TheDZhon (approved):
:green_apple:  ty ty!

PR #35, opened 2024-02-07
Cfg: wstETH on Scroll

TheDZhon:
- [x] Add addresses for the proposed deployment on Scroll Sepolia: https://github.com/lidofinance/docs/pull/323
- [x] Fix logging for GitHub fetch errors

### DIFF
--- a/config_samples/scroll/mainnet/scroll_mainnet_config_L1.json
+++ b/config_samples/scroll/mainnet/scroll_mainnet_config_L1.json
@@ -1,0 +1,28 @@
+{
+    "contracts": {
+        "0x6625c6332c9f91f2d27c304e729b86db87a3f504": "TransparentUpgradeableProxy",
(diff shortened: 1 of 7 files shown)


TheDZhon:
There is a little diff :thinking: 
[IL1GatewayRouter.sol.html.pdf](https://github.com/lidofinance/diffyscan/files/14361891/IL1GatewayRouter.sol.html.pdf)

arwer13:
And the diff is in the interface of an externally called contract. In `calldata` instead of `memory` annotator of function arg so the signature of the call isn't changed. And the bytecode of the deployment is not affected.

But quite unexpected indeed.

arwer13 (approved):
LGTM: addresses match the ones in:
- https://github.com/lidofinance/docs/pull/323
- https://github.com/lidofinance/periphery/blob/l2-acceptance-script/scripts/acceptance-tests/scroll/scroll-mainnet.yaml

joao1963:
92928b31-f9cf-4e6f-a6e5-3266d9c5d545

PR #47, opened 2024-06-03
build: make installable via poetry / pipx

madlabman:


mymphe (approved):
Awesome feature!

pyproject.toml
```diff
@@ -1,22 +1,25 @@
 [tool.poetry]
-name = "explorer-github-diff-checker"
-version = "0.2.0"
+name = "diffyscan"
+version = "0.1.0"
```
TheDZhon, line 3: Sorry, but if we want to package it, then I would plug the automated versioned builds here:

```toml
version = "0.0.0"

# ...

[tool.poetry-dynamic-versioning]
enable = true
vcs = "git"
style = "semver"
metadata = false

[build-system]
requires = ["poetry-core>=1.0.0", "poetry-dynamic-versioning>=1.1.0,<2.0.0"]
build-backend = "poetry_dynamic_versioning.backend"
```

TheDZhon: Also, please consider adding a command line argument to retrieve the version upon launch.

TheDZhon (changes requested):
Thank you for the great proposal!

I'd suggest addressing only a single suggestion regarding the version bookkeeping approach to make it more robust.

skhomuti (approved):

TheDZhon (approved):
Awesome work, merging to the upstream :+1:

Issue #13, opened 2023-07-14
Proposed enhancement: Integrate Bytecode Verificator into Diffyscan


TheDZhon:
In light of the recent deployment of Lido V2, we have significantly augmented our deployments by introducing two versatile tools:

- Diffyscan: A tool designed by @mymphe to cross-verify our source code with the version published on Etherscan.
- Bytecode Verificator: A concise bash-script-based tool crafted by @bulbozaur. You can find this utility in the `lido-dao` repository [here](https://github.com/lidofinance/lido-dao/tree/master/bytecode-verificator).

As we continue to evolve our development practices, consider integrating the functionality of the Bytecode Verificator directly into the Diffyscan module.

![image](https://github.com/lidofinance/diffyscan/assets/961317/6cf1d4b4-4e34-4c0e-9b95-0dd9a3767ab3)

TheDZhon:
Available in the https://github.com/lidofinance/diffyscan/releases/tag/v0.10.0 release

Issue #10, opened 2023-05-03
Commentable config format


folkyatina:
It would be lovely to have a possibility to have some comments in config, but JSON does not support it by default. 

I think we should consider some other format for config:
- toml
- yaml
- raw .py files
- something else

TheDZhon:
JYC: Starting from 3.11, Python has the built-in `tomlib`.

TheDZhon:
implemented yaml

PR #52, opened 2024-07-07
Add Ethereum Smart Contract binary bytecode matching

Alexvozhak:
**Description**:

This PR enhances `diffyscan` by integrating Ethereum smart contract binary matching. It incorporates functionalities from [verify-bytecode.ts](https://github.com/lidofinance/stonks/blob/main/scripts/verify-bytecode.ts) and [bytecode_verificator.sh](https://github.com/lidofinance/lido-dao/blob/master/bytecode-verificator/bytecode_verificator.sh).

---

**Key Changes**:

- **Showcases**:
  - **Sepolia Network**: Added support for all contracts by extending [`lido_dao_sepolia_config.json`](https://github.com/lidofinance/diffyscan/blob/main/config_samples/lido_dao_sepolia_config.json).
  - **Holesky Network**: Added support for 3 out of 53 contracts by extending [`lido_dao_holesky_config.json`](https://github.com/lidofinance/diffyscan/blob/main/config_samples/lido_dao_holesky_config.json).

- **New Modules and Refactoring**:
  - **Binary Comparison Logic**: Introduced `binary_verifier.py` to handle binary comparison of contracts.
  - **Argument Encoding**: Added `encoder.py` for encoding contract arguments, referencing the [Solidity ABI specification](https://docs.soliditylang.org/en/develop/abi-spec.html).
  - **Compiler Module**: Moved all source compilation code to `compiler.py`.
  - **Calldata Handling**: Created `calldata.py` for calculating and parsing constructor arguments from configurations.
  - **Type Definitions**: Renamed and extended `types.py` to `custom_types.py`.
  - **Node Interactions**:
    - Encapsulated interactions with the local Hardhat node in `hardhat.py`.
    - Moved common node operations to `node_handler.py`.
  - **Exception Handling**: Added specific exceptions for new features in `custom_exceptions.py`.

- **Configurations and Documentation**:
  - **Configuration Updates**:
    - Extended `lido_dao_sepolia_config.json` and `lido_dao_holesky_config.json` with a `binary_comparison` section.
    - Added specific Hardhat configurations for Sepolia and Holesky networks.
  - **Documentation**:
    - Updated the README to reflect new features and usage instructions.
    - Included support for user-defined Hardhat configurations as shown in the README example.

- **Improvements and Fixes**:
  - **Code Quality**:
    - Fixed existing errors identified by the `mypy` linter.
    - Enhanced code readability and organization.
  - **Backwards compatibility**: Ensured that existing `diffyscan` usage remains unaffected.

---

**Implemented Logic for Contract Code Matching**:

1. **Source Retrieval**: Obtain contract sources from Etherscan.
2. **Source Diffing**: Perform source code diffing as previously implemented.
3. **Bytecode Matching**: Compile the contracts and attempt to match the bytecode obtained from the RPC node.
4. **Matching Outcome**:
   - **Success**: If the bytecode matches, proceed to the next contract.
   - **Metadata Differences**: If only the metadata differs, display a warning but continue processing.
5. **Error Handling**:
   - **No Immutables**: If there are no immutables and the bytecode doesn't match, display an error.
   - **With Immutables**: If immutables are present and the configuration includes constructor calldata or parameters, attempt to deploy the contract and compare the resulting bytecode.

---

By integrating these enhancements, `diffyscan` has been enriched with the ability to verify smart contracts at the bytecode level, ensuring greater reliability and security in contract deployment and verification processes.


README.md
tamtamchik, line 1: Locking on ganache (outdated and even archived https://github.com/trufflesuite/ganache) looks like a major issue. I think a node-agnostic setup with an anvil/hardhat node running on a specific port will be better.

tamtamchik: I have a request: can you please format the code in `Black` style so it's easier to read?

diffyscan/utils/binary_verifier.py
```diff
@@ -0,0 +1,333 @@
...
+    os.chmod(compiler_path, compiler_file_rights.st_mode | stat.S_IEXEC)
+    
+def compile_contracts(compiler_path, input_settings):
+    process = subprocess.run([compiler_path,'--standard-json'], input=input_settings.encode(), capture_output=True, check=True, timeout=30)
+    return json.loads(process.stdout)
+
+def prepare_compiler(platform, build_info, compiler_path):
+    create_dirs(compiler_path)
+    compiler_binary = download_compiler(platform, build_info, compiler_path)
+    valid_checksum = build_info['sha256'][2:];
```
tamtamchik, line 57:
```suggestion
    valid_checksum = build_info['sha256'][2:]
```

diffyscan/utils/binary_verifier.py
```diff
@@ -0,0 +1,333 @@
...
+  if response_getTransactionReceipt['result']['status'] != '0x1':
+    logger.error(f'Failed to received transaction receipt. Transaction has been reverted (status 0x0). Input missmatch?')
+    return None
+  
+  contract_address = response_getTransactionReceipt['result']['contractAddress']
+  
+  return contract_address
+    
+          
+OPCODES = {
```
tamtamchik, line 193: Move to constants?

diffyscan/utils/common.py
```diff
@@ -43,7 +40,21 @@ def fetch(url, headers={}):
         logger.error("Status", response.status_code)
         logger.error("Response", response.text)
 
-    return response.json()
+    return response
+
+
+def fetch(url, headers={}):
```
tamtamchik, line 46:
```suggestion
def fetch(url, headers=None):
```

diffyscan/utils/common.py
```diff
@@ -43,7 +40,21 @@ def fetch(url, headers={}):
...
+    logger.log(f"Fetch: {url}")
+    response = requests.get(url, headers=headers)
+
+    return handle_response(response, url)
+
+
+def pull(url, payload={}):
```
tamtamchik, line 53:
```suggestion
def pull(url, payload=None):
```

diffyscan/utils/common.py
```diff
@@ -57,3 +68,13 @@ def parse_repo_link(repo_link):
...
     user_slash_repo = repo_location[0]
     return user_slash_repo
+
+
+def get_solc_native_platform_from_os():
+    platform_name = sys.platform
+    if platform_name == "linux":
+        return "linux-amd64"
+    elif platform_name == "darwin":
+        return "macosx-amd64"
```
tamtamchik, line 78: It can also be `macosx-arm64` (for Apple M chip based)

diffyscan/utils/binary_verifier.py
```diff
@@ -0,0 +1,333 @@
...
+
+    for ind in list(checkpoints):
+        startIndex = max(0, ind - nearLinesCount)
+        lastIndex = min(ind + nearLinesCount, maxInstructionsCount - 1)
+        for i in range(startIndex, lastIndex + 1):
+            checkpoints.add(i)
+
+    checkpointsArray = sorted(list(checkpoints))
+
+    def hex(index, padStart=2):
```
tamtamchik, line 265: Move to logger?

tamtamchik (changes requested):
Round 2: Added a bunch of new comments and suggestions. 

Run the testing command: 
```
diffyscan config_samples/lido_dao_sepolia_config.json hardhat_configs/sepolia_hardhat.config.js --binary-check -y
```
And I see some errors in logs:

```
🔵 [INFO] Receiving the account from "http://127.0.0.1:7545" ...
 🟢 [OKAY] The account was successfully received
 🔵 [INFO] Started binary checking for 0x8A1AA86d35b2EE8C9369618E7D7b40000cCD3295
 🟢 [OKAY] Contracts were successfully compiled. The target contract is ready for matching
 🔵 [INFO] Constructor args types: ['address', 'bytes32', 'bytes']
 🔵 [INFO] Trying to deploy transaction to "http://127.0.0.1:7545" ...
 🔴 [ERROR] Failed to deploy transaction: Error: Transaction reverted without a reason string
 🔴 [ERROR] Failed to deploy bytecode to http://127.0.0.1:7545, binary diff skipped (0x8A1AA86d35b2EE8C9369618E7D7b40000cCD3295)```
 
I think such errors should not be skipped, and checking should be stopped on such errors.

File: diffyscan/diffyscan.py [hunk 7]
```diff
@@ -7,35 +7,104 @@
...
+from .utils.binary_verifier import *
+from .utils.hardhat import hardhat
 
 __version__ = "0.0.0"
 
 
 g_skip_user_input: bool = False
 
+
 def prettify_solidity(solidity_contract_content: str):
```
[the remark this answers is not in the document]
Alexvozhak: done

diffyscan/utils/constants.py
```diff
@@ -1,8 +1,25 @@
...
 
 DIGEST_DIR = "digest"
 START_TIME = time.time()
 START_TIME_INT = int(START_TIME)
 DIFFS_DIR = f"{DIGEST_DIR}/{START_TIME_INT}/diffs"
 LOGS_PATH = f"{DIGEST_DIR}/{START_TIME_INT}/logs.txt"
 DEFAULT_CONFIG_PATH = "config.json"
+
+REMOTE_RPC_URL = os.getenv("REMOTE_RPC_URL", "")
+if not REMOTE_RPC_URL:
```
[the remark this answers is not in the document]
Alexvozhak: done

diffyscan/utils/constants.py
```diff
@@ -1,8 +1,25 @@
...
+    raise ValueError("REMOTE_RPC_URL variable is not set")
+
+SOLC_DIR = os.getenv("SOLC_DIR", "")
+if SOLC_DIR == "":
```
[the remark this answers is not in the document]
Alexvozhak: done

diffyscan/utils/constants.py
```diff
@@ -1,8 +1,25 @@
...
+    raise ValueError("SOLC_DIR variable is not set")
+
+LOCAL_RPC_URL = os.getenv("LOCAL_RPC_URL", "")
```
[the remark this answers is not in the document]
Alexvozhak: done

README.md
```diff
@@ -34,11 +35,22 @@ Set your Github token to query API without strict rate limiting,
...
+Set local and remote RPC URLs for matching bytecode from local node againts remote
+
+```bash
+export REMOTE_RPC_URL=<remote-rpc-url> //https://ethereum-sepolia-rpc.publicnode.com
+export LOCAL_RPC_URL=<local-rpc-url> //http://127.0.0.1:7545
+```
+Set absolute or relative path for downloading solc compilers
+
+```bash
+export SOLC_DIR=<solc-path>
```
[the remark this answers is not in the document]
Alexvozhak: done

README.md
```diff
@@ -5,17 +5,18 @@
...
+Supports binary matching deployed bytecode from local node (Hardhat) against remote (option `--binary-check`)
 
 ## Install
 
 ```bash
 pipx install git+https://github.com/lidofinance/diffyscan
 ```
 
-If need `--prettify` option
+If need `--prettify`, '--binary-check' options
```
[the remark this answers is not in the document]
Alexvozhak: done

diffyscan/utils/binary_verifier.py
```diff
@@ -0,0 +1,333 @@
...
+from .helpers import create_dirs
+from .logger import logger
+from .encoder import encode_constructor_arguments
+from .constants import SOLC_DIR
+  
+def get_compiler_info(platform, required_compiler_version):
+    compilers_list_url = (
+      f'https://raw.githubusercontent.com/ethereum/solc-bin/gh-pages/{platform}/list.json'   
+    )
+    available_compilers_list = fetch(compilers_list_url).json()
```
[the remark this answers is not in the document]
Alexvozhak: done

diffyscan/utils/binary_verifier.py
```diff
@@ -0,0 +1,333 @@
...
+        shutil.rmtree(SOLC_DIR)
+        logger.okay(f"{SOLC_DIR} deleted")
+
+def download_compiler(platform, build_info, destination_path):
+    compiler_url = (
+                    f'https://binaries.soliditylang.org/{platform}/{build_info["path"]}'
+    )
+    download_compiler_response = fetch(compiler_url)  
+
+    with open(destination_path, 'wb') as compiler_file:
```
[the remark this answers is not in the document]
Alexvozhak: done

diffyscan/utils/binary_verifier.py
```diff
@@ -0,0 +1,333 @@
...
+    compiler_checksum = hashlib.sha256(compiler).hexdigest()
+    if compiler_checksum != valid_checksum:
+      raise ValueError(f'Bad checksum')
+    
+def set_compiler_executable(compiler_path):
+    compiler_file_rights = os.stat(compiler_path)
+    os.chmod(compiler_path, compiler_file_rights.st_mode | stat.S_IEXEC)
+    
+def compile_contracts(compiler_path, input_settings):
+    process = subprocess.run([compiler_path,'--standard-json'], input=input_settings.encode(), capture_output=True, check=True, timeout=30)
```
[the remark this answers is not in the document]
Alexvozhak: done

diffyscan/utils/binary_verifier.py
Alexvozhak: done

diffyscan/utils/binary_verifier.py
Alexvozhak: done

diffyscan/utils/common.py
Alexvozhak: done

diffyscan/utils/common.py
Alexvozhak: done

diffyscan/diffyscan.py
```diff
@@ -145,32 +213,94 @@ def run_diff(config, name, address, explorer_api_token, github_api_token, recurs
...
+    parser.add_argument(
+        "--prettify",
+        "-p",
+        help="Unify formatting by prettier before comparing",
+        action="store_true",
+    )
+    parser.add_argument(
+        "--binary-check",
+        "-b",
+        help="Match contracts by binaries such as verify-bytecode.ts",
```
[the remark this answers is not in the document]
Alexvozhak: done

diffyscan/utils/binary_verifier.py
```diff
@@ -0,0 +1,333 @@
...
+    download_compiler_response = fetch(compiler_url)  
+
+    with open(destination_path, 'wb') as compiler_file:
+      compiler_file.write(download_compiler_response.content)
+    return download_compiler_response.content
+      
+def check_compiler_checksum(compiler, valid_checksum):
+    compiler_checksum = hashlib.sha256(compiler).hexdigest()
+    if compiler_checksum != valid_checksum:
+      raise ValueError(f'Bad checksum')
```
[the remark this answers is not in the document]
Alexvozhak: done

diffyscan/utils/binary_verifier.py
Alexvozhak: done

diffyscan/utils/binary_verifier.py
```diff
@@ -0,0 +1,333 @@
...
+  
+def get_compiler_info(platform, required_compiler_version):
+    compilers_list_url = (
+      f'https://raw.githubusercontent.com/ethereum/solc-bin/gh-pages/{platform}/list.json'   
+    )
+    available_compilers_list = fetch(compilers_list_url).json()
+    required_build_info = next((compiler for compiler in available_compilers_list['builds'] if compiler['longVersion'] == required_compiler_version), None)
+
+    if not required_build_info:
+      raise ValueError(f'Required compiler version for "{platform}" is not found')
```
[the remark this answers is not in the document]
Alexvozhak: For example?

diffyscan/utils/binary_verifier.py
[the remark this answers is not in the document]
Alexvozhak: done

README.md
```diff
@@ -34,11 +35,22 @@ Set your Github token to query API without strict rate limiting,
...
+
+```bash
+export SOLC_DIR=<solc-path>
+```
 
 Start script with one of the examples provided (or entire folder of configs)
 
 ```bash
-diffyscan config_samples/lido_dao_sepolia_config.json
+diffyscan config_samples/lido_dao_sepolia_config.json hardhat_configs/sepolia_hardhat.config.js
```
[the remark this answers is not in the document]
Alexvozhak: I'm not sure that I correctly understood how it should work

diffyscan/utils/hardhat.py
```diff
@@ -0,0 +1,77 @@
...
+        hardhat_config_relative_path = Hardhat.get_config_path(
+            os.path.dirname(main_config_relative_path),
+            "hardhat_configs",
+            binary_config["hardhat_config_name"],
+        )
+        local_node_command = (
+            f"npx hardhat node --hostname {parsed_url.hostname} "
+            f"--port {parsed_url.port} "
+            f"--config {hardhat_config_relative_path} "
+            f"--fork {binary_config["remote_RPC_URL"]}"
```
Alexvozhak, line 33: @tamtamchik after adding sepolia fork deploy stopped to work, this problem hasn't solved yet

Alexvozhak:
@tamtamchik hardhat needs to be delivered manually (npm install hardhat), i will try to fix it

diffyscan/utils/hardhat.py
tamtamchik: @Alexvozhak thank you for the update

README.md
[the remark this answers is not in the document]
tamtamchik: Basically, you can have one `hardhat.config.js` file that will be automatically used by hardhat and setup all the networks there (example, https://github.com/lidofinance/core/blob/master/hardhat.config.ts#L40-L68), then just specify `--network sepolia` on node start and HH will automatically use config from this file.

diffyscan/utils/binary_verifier.py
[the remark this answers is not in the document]
tamtamchik: For this example, it may look something like this. This is a "nice to have" feature since `diffyscan,` I assume, can be used as a module and therefore should produce meaningful error types, not generic ones.  Disclaimer: I might be wrong, and this is not in the scope. cc @TheDZhon 

```python
class CompilerVersionNotFoundError(Exception):
    def __init__(self, platform, version):
        self.platform = platform
        self.version = version
        self.message = f'Required compiler version "{version}" for platform "{platform}" is not found'
        super().__init__(self.message)
        
...
raise CompilerVersionNotFoundError(platform, required_compiler_version)
```

README.md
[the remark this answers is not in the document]
Alexvozhak: It won't work like that, [proof](https://github.com/NomicFoundation/hardhat/issues/2449#issuecomment-1059259560)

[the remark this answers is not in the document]
tamtamchik: @Alexvozhak roger that! 🫡

Alexvozhak:
@tamtamchik hardhat autoinstall has been fixed

tamtamchik:
JFYI, I have changed https://ethereum-sepolia-rpc.publicnode.com to https://sepolia.drpc.org, and now it runs OK at least for Sepolia.

Alexvozhak:
@tamtamchik i have already updated PR with new URL, it really works!

> JFYI, I have changed https://ethereum-sepolia-rpc.publicnode.com to https://sepolia.drpc.org, and now it runs OK at least for Sepolia, no errors.
> 
> I plan to review it again tomorrow and probably approve it if no showstoppers are found. @Alexvozhak, great work! Thank you!

README.md
```diff
@@ -5,17 +5,18 @@
 ![license MIT](https://img.shields.io/badge/license-MIT-brightgreen)
 [![Code style: black](https://img.shields.io/badge/code%20style-black-000000.svg)](https://github.com/psf/black)
 
-Diff your Ethereum smart contracts code from GitHub against Blockchain explorer verified source code.
+Diff your Ethereum smart contracts code from GitHub against Blockchain explorer verified source code, deployed bytecode from local node (Hardhat) against remote
```
tamtamchik, line 8: Looks unfinished.

README.md
```diff
@@ -5,17 +5,18 @@
...
 
 Supports reformatting solidity code by means of prettifier solidity plugin before comparing the sources (option `--prettify`).
+Supports binary checking deployed bytecode from local node (Hardhat) against remote (see section 'binary_checking' in lido_dao_sepolia_config.json as example)
```
tamtamchik, line 11:
```suggestion
Supports binary checking deployed bytecode from the local node (Hardhat) against remote (see section 'binary_checking' in `./config_samples/lido_dao_sepolia_config.json` as an example)
```

README.md
```diff
@@ -5,17 +5,18 @@
...
 
 ## Install
 
 ```bash
 pipx install git+https://github.com/lidofinance/diffyscan
 ```
 
-If need `--prettify` option
+If need `--prettify` or binary checking options support
```
tamtamchik, line 19:
```suggestion
If need `--prettify` or binary checking options support:
```

README.md
```diff
@@ -90,7 +91,7 @@ This project was developed using these dependencies with their exact versions li
 
 - Python 3.12
 - Poetry 1.8
-- if need `--prettify` option support:
+- if need `--prettify` or binary checking options support
```
tamtamchik, line 94:
```suggestion
- if need `--prettify` or binary checking options support:
```

README.md
```diff
@@ -128,7 +129,7 @@ poetry shell
 poetry install
 ```
 
-5. If need `--prettify` option
+5. If need `--prettify` or binary checking options support
```
tamtamchik, line 132:
```suggestion
5. If need `--prettify` or binary checking options support:
```

config_samples/lido_dao_holesky_config.json
```diff
@@ -111,5 +112,38 @@
             "commit": "6bd6b76d1156e20e45d1016f355d154141c7e5b9",
             "relative_root": "contracts"
         }
+    },
+    "binary_checking": {
+	"use": 1,
+	"skip_deploy_error" : 1,
```
tamtamchik, line 118: It should be false by default, I think.

config_samples/lido_dao_holesky_config.json
```diff
@@ -111,5 +112,38 @@
             "commit": "6bd6b76d1156e20e45d1016f355d154141c7e5b9",
             "relative_root": "contracts"
         }
+    },
+    "binary_checking": {
+	"use": 1,
```
tamtamchik, line 117:
```suggestion
	"enable": true,
```

config_samples/lido_dao_holesky_config.json
tamtamchik, line 118:
```suggestion
	"skip_deploy_error" : false,
```

diffyscan/diffyscan.py
```diff
@@ -149,28 +217,90 @@ def process_config(path: str, recursive_parsing: bool, unify_formatting: bool):
     logger.info(f"Loading config {path}...")
     config = load_config(path)
 
-    github_api_token = load_env("GITHUB_API_TOKEN", masked=True)
     explorer_token = None
     if "explorer_token_env_var" in config:
-        explorer_token = load_env(config["explorer_token_env_var"], masked=True, required=False)
+        explorer_token = load_env(
+            config["explorer_token_env_var"], masked=True, required=False
```
tamtamchik, line 223: All new config values should be properly typed and added to `utils→types→Config`

diffyscan/utils/compiler.py
```diff
@@ -0,0 +1,109 @@
...
+    platform_name = sys.platform
+    if platform_name == "linux":
+        return "linux-amd64"
+    elif platform_name == "darwin":
+        return "macosx-amd64" if platform.machine() == "x86_64" else "macosx-arm64"
+    else:
+        raise ValueError(f"Unsupported platform {platform_name}")
+
+
+def get_compiler_info(platform, required_compiler_version):
```
tamtamchik, line 24: platform shadows the import platform 
```suggestion
def get_compiler_info(required_platform, required_compiler_version):
```

diffyscan/utils/compiler.py
```diff
@@ -0,0 +1,109 @@
...
+    compilers_list_url = f"https://raw.githubusercontent.com/ethereum/solc-bin/gh-pages/{platform}/list.json"
```
tamtamchik, line 25:
```suggestion
    compilers_list_url = f"https://raw.githubusercontent.com/ethereum/solc-bin/gh-pages/{required_platform}/list.json"
```

diffyscan/utils/compiler.py
```diff
@@ -0,0 +1,109 @@
...
+        (
+            compiler
+            for compiler in available_compilers_list["builds"]
+            if compiler["longVersion"] == required_compiler_version
+        ),
+        None,
+    )
+
+    if not required_build_info:
+        raise ValueError(f'Required compiler version for "{platform}" is not found')
```
tamtamchik, line 37:
```suggestion
        raise ValueError(f'Required compiler version for "{required_platform}" is not found')
```

diffyscan/utils/compiler.py
```diff
@@ -0,0 +1,109 @@
...
+
+    return required_build_info
+
+
+def download_compiler(platform, build_info, destination_path):
```
tamtamchik, line 42:
```suggestion
def download_compiler(required_platform, build_info, destination_path):
```

diffyscan/utils/compiler.py
```diff
@@ -0,0 +1,109 @@
...
+    compiler_url = f'https://binaries.soliditylang.org/{platform}/{build_info["path"]}'
```
tamtamchik, line 43:
```suggestion
    compiler_url = f'https://binaries.soliditylang.org/{required_platform}/{build_info["path"]}'
```

diffyscan/utils/compiler.py
```diff
@@ -0,0 +1,109 @@
...
+            f"Compiler checksum mismatch. Expected: {valid_checksum}, Got: {compiler_checksum}"
+        )
+
+
+def set_compiler_executable(compiler_path):
+    compiler_file_rights = os.stat(compiler_path)
+    os.chmod(compiler_path, compiler_file_rights.st_mode | stat.S_IEXEC)
+
+
+def prepare_compiler(platform, build_info, compiler_path):
```
tamtamchik, line 69:
```suggestion
def prepare_compiler(required_platform, build_info, compiler_path):
```

diffyscan/utils/compiler.py
```diff
@@ -0,0 +1,109 @@
...
+    create_dirs(compiler_path)
+    compiler_binary = download_compiler(platform, build_info, compiler_path)
```
tamtamchik, line 71:
```suggestion
    compiler_binary = download_compiler(required_platform, build_info, compiler_path)
```

diffyscan/utils/encoder.py
```diff
@@ -0,0 +1,109 @@
...
+            encoded_offsets += encode_uint256(arg_value)
+        elif arg_type == "bytes32":
+            encoded_offsets += encode_bytes32(arg_value)
+        elif arg_type == "address[]" and not arg_value:
+            encoded_data += "0" * 64
+            offset = format((arg_index + args_length) * 32, "064x")
+            encoded_offsets += offset
+        else:
+            logger.warn(
+                f"Unknown constructor argument type '{arg_type}', use --constructor-calldata instead"
```
tamtamchik, line 47: I can't find any `--constructor-calldata` except this place, how should this work?

diffyscan/utils/explorer.py
```diff
@@ -55,16 +79,16 @@ def _get_contract_from_zksync(zksync_explorer_hostname, contract):
...
+        "sources": json.loads(data["sourceCode"]["sources"]),
+        "compiler": data["CompilerVersion"],
+    }
 
 
 def _get_contract_from_mantle(mantle_explorer_hostname, contract):
     etherscan_link = f"https://{mantle_explorer_hostname}/api?module=contract&action=getsourcecode&address={contract}"
-
-    response = fetch(etherscan_link)
+    response = fetch(etherscan_link).json
```
tamtamchik, line 91:
```suggestion
    response = fetch(etherscan_link).json()
```

diffyscan/utils/explorer.py
```diff
@@ -15,36 +17,58 @@ def _get_contract_from_etherscan(token, etherscan_hostname, contract):
...
+        }
 
 
 def _get_contract_from_zksync(zksync_explorer_hostname, contract):
     zksync_explorer_link = (
         f"https://{zksync_explorer_hostname}/contract_verification/info/{contract}"
     )
 
-    response = fetch(zksync_explorer_link)
+    response = fetch(zksync_explorer_link).json
```
tamtamchik, line 71:
```suggestion
    response = fetch(zksync_explorer_link).json()
```

diffyscan/utils/explorer.py
```diff
@@ -15,36 +17,58 @@ def _get_contract_from_etherscan(token, etherscan_hostname, contract):
...
-        json.loads(data["SourceCode"][1:-1])["sources"].items()
-        if json_escaped
-        else json.loads(data["SourceCode"]).items()
-    )
-
-    return (contract_name, source_files)
+    solc_input = result["SourceCode"]
+
+    if solc_input.startswith("{{"):
+        return {
```
tamtamchik, line 32: Can we add a returning type to all such functions so it'll be easier to see that the return value is valid?

diffyscan/utils/hardhat.py
```diff
@@ -0,0 +1,77 @@
...
+                logger.okay(f"Hardhat successfully started, PID {self.sub_process.pid}")
+            else:
+                raise ValueError(f"Failed to start Hardhat: something is wrong")
+
+    def stop(self):
+        if self.sub_process is not None and self.sub_process.poll() is None:
+            os.kill(self.sub_process.pid, signal.SIGTERM)
+            logger.info(f"Hardhat stopped, PID {self.sub_process.pid}")
+
+    def get_config_path(from_path: str, to_path: str, filename: str) -> str:
```
tamtamchik, line 70: `@staticmethod`?

tamtamchik (commented):
All in all looks good, but I got an error again: 

`diffyscan ./config_samples/lido_dao_holesky_config.json`

```
 🔵 [INFO] Started binary checking for (0x3F1c547b21f65e10480dE3ad8E19fAAC46C95034 : AppProxyUpgradeable)
 🟢 [OKAY] Contracts were successfully compiled. The target contract is ready for matching
 🔵 [INFO] Hardhat stopped, PID 20053
Traceback (most recent call last):
  File ".../diffyscan-9kJxk0py-py3.12/bin/diffyscan", line 6, in <module>
    sys.exit(main())
             ^^^^^^
  File ".../diffyscan-fork/diffyscan/diffyscan.py", line 316, in main
    process_config(args.path, args.support_brownie, args.prettify)
  File ".../diffyscan-fork/diffyscan/diffyscan.py", line 261, in process_config
    run_binary_diff(
  File ".../diffyscan-fork/diffyscan/diffyscan.py", line 39, in run_binary_diff
    contract_creation_code, immutables = get_code_from_explorer(
                                         ^^^^^^^^^^^^^^^^^^^^^^^
  File ".../diffyscan-fork/diffyscan/utils/explorer.py", line 186, in get_code_from_explorer

ValueError: Failed to find constructor args values 0x3F1c547b21f65e10480dE3ad8E19fAAC46C95034
```

config_samples/lido_dao_holesky_config.json
tamtamchik: Should be handled in the code accordingly.

diffyscan/utils/hardhat.py
Alexvozhak: done

README.md
Alexvozhak: done

Alexvozhak:
> All in all looks good, but I got an error again:
> 
> `diffyscan ./config_samples/lido_dao_holesky_config.json`
> 
> ```
>  🔵 [INFO] Started binary checking for (0x3F1c547b21f65e10480dE3ad8E19fAAC46C95034 : AppProxyUpgradeable)
>  🟢 [OKAY] Contracts were successfully compiled. The target contract is ready for matching
>  🔵 [INFO] Hardhat stopped, PID 20053
> Traceback (most recent call last):
>   File ".../diffyscan-9kJxk0py-py3.12/bin/diffyscan", line 6, in <module>
>     sys.exit(main())
>              ^^^^^^
>   File ".../diffyscan-fork/diffyscan/diffyscan.py", line 316, in main
>     process_config(args.path, args.support_brownie, args.prettify)
>   File ".../diffyscan-fork/diffyscan/diffyscan.py", line 261, in process_config
>     run_binary_diff(
>   File ".../diffyscan-fork/diffyscan/diffyscan.py", line 39, in run_binary_diff
>     contract_creation_code, immutables = get_code_from_explorer(
>                                          ^^^^^^^^^^^^^^^^^^^^^^^
>   File ".../diffyscan-fork/diffyscan/utils/explorer.py", line 186, in get_code_from_explorer
> 
> ValueError: Failed to find constructor args values 0x3F1c547b21f65e10480dE3ad8E19fAAC46C95034
> ```

Error, because now only 3 contracts from 53 are filled in holesky config (constructor's values)

config_samples/lido_dao_holesky_config.json
Alexvozhak: done

config_samples/lido_dao_holesky_config.json
Alexvozhak: done

config_samples/lido_dao_holesky_config.json
Alexvozhak: done

diffyscan/diffyscan.py
Alexvozhak: I don't understand what role it plays. If I comment it out or if I add incorrect types, nothing breaks. 
So I suggest to remove it completely

diffyscan/utils/compiler.py
Alexvozhak: done

diffyscan/utils/compiler.py
Alexvozhak: done

diffyscan/utils/compiler.py
Alexvozhak: done

diffyscan/utils/compiler.py
Alexvozhak: done

diffyscan/utils/compiler.py
Alexvozhak: done

diffyscan/utils/encoder.py
Alexvozhak: done. Added field 'constructor_calldata' into config for using prepared calldata, see updates into sepolia and holesky configs

tamtamchik (approved):
Don't have significant questions or notices. Looks good to me. Approved.

diffyscan/utils/binary_verifier.py
```diff
@@ -0,0 +1,142 @@
...
+            raise ValueError("Invalid bytecode difference data")
+
+
+def parse(bytecode):
+    buffer = bytes.fromhex(bytecode[2:] if bytecode.startswith("0x") else bytecode)
+    instructions = []
+    i = 0
+    while i < len(buffer):
+        opcode = buffer[i]
+        length = 1 + (opcode - 0x5F if 0x5F <= opcode <= 0x7F else 0)
```
TheDZhon, line 132: why `0x5F` and `0x7F` aren't constants?

diffyscan/utils/binary_verifier.py
```diff
@@ -0,0 +1,142 @@
...
+                else bgRed(actual["op"]["name"]) + " " + bgGreen(expected["op"]["name"])
+            )
+            actualParams = (
+                "0x" + actual["bytecode"][2:] if len(actual["bytecode"]) > 2 else ""
+            )
+            expectedParams = (
+                "0x" + expected["bytecode"][2:] if len(expected["bytecode"]) > 2 else ""
+            )
+
+            paramsLength = len(expected["bytecode"]) // 2 - 1
```
TheDZhon, line 110: It seems here we deal with a hex string `0xAABBCCDD` and try to count the bytes inside.
shall it be done via a little helper func to be more reader-friendly?

diffyscan/utils/calldata.py
```diff
@@ -0,0 +1,45 @@
+from .logger import logger
+from .encoder import encode_constructor_arguments
+
+def get_calldata(contract_address_from_config, target_compiled_contract, binary_config):
+    calldata = get_prepared_calldata_from_config(contract_address_from_config, binary_config)
+    if calldata is not None:
+          return calldata, ''
```
TheDZhon, line 7:
```suggestion
          return calldata, None
```
Here, we have a function that returns `calldata` and error (optionally). 

Why return empty text instead of `None` in such a case?

diffyscan/utils/calldata.py
```diff
@@ -0,0 +1,45 @@
...
+    
+    calldata, text_error = parse_calldata_from_config(
+        contract_address_from_config,
+        binary_config["constructor_args"],
+        target_compiled_contract,
+    )
+    return calldata, text_error
```
TheDZhon, line 14: Thinking further, I am a bit puzzled why we return `text_error` and `calldata` together.
It seems that `text_error` might arise only from the parsed config stage. 

Should it be raised as an exception instead?

diffyscan/utils/calldata.py
```diff
@@ -0,0 +1,45 @@
...
+            
+def get_constructor_abi(target_compiled_contract):
+    constructor_abi = None
+    try:
+        constructor_abi = [entry["inputs"] for entry in target_compiled_contract['abi'] if entry["type"] == "constructor"][0]
```
TheDZhon, line 19: looks like this file hasn't been formatted with black, the line is massively long

diffyscan/utils/calldata.py
```diff
@@ -0,0 +1,45 @@
...
+    except IndexError:
+        logger.okay(f'Constructor in ABI not found, calldata calculation skipped')
```
TheDZhon, line 21:
```suggestion
        logger.okay(f'Contract's ABI doesn't have a constructor, calldata calculation skipped')
```

diffyscan/utils/calldata.py
```diff
@@ -0,0 +1,45 @@
...
+        return None
+
+    logger.okay(f'Constructor in ABI successfully found: {[arg['type'] for arg in constructor_abi]}')
```
TheDZhon, line 24: might be too verbose to log every step

diffyscan/utils/calldata.py
```diff
@@ -0,0 +1,45 @@
...
+    return constructor_abi
+
+def get_prepared_calldata_from_config(contract_address_from_config, binary_config):
```
TheDZhon, line 27:
```suggestion
def get_raw_calldata_from_config(contract_address_from_config, binary_config):
```

diffyscan/utils/calldata.py
```diff
@@ -0,0 +1,45 @@
...
+    if "constructor_calldata" not in binary_config or contract_address_from_config not in binary_config["constructor_calldata"]:
+        return None
+    calldata_field = binary_config["constructor_calldata"]
+    logger.info(f"Trying to use prepared calldata from config")
+    prepared_calldata_from_config = calldata_field[contract_address_from_config]
+    return prepared_calldata_from_config
+    
+def parse_calldata_from_config(contract_address_from_config, constructor_args, target_compiled_contract):
```
TheDZhon, line 35: If you return composed types, consider providing type hints for the function declarations, e.g.:

```python
def parse_calldata_from_config(
    contract_address_from_config, 
    constructor_args, 
    target_compiled_contract
) → (Optional[str], Optional[str]):
```

TheDZhon: Nevertheless, I am not sure that we should use Optional types and not raising exceptions

The file feels like a mix of both approaches, but I may miss an intended structure

diffyscan/utils/common.py
```diff
@@ -31,19 +33,38 @@ def load_config(path: str) -> Config:
...
+        response = requests.get(url, headers=headers)
+        response.raise_for_status()
+    except requests.exceptions.HTTPError as http_err:
+        raise ValueError(f"HTTP error occurred: {http_err}")
+    except requests.exceptions.ConnectionError as conn_err:
+        raise ValueError(f"Connection error occurred: {conn_err}")
+    except requests.exceptions.Timeout as timeout_err:
+        raise ValueError(f"Timeout error occurred: {timeout_err}")
+    except requests.exceptions.RequestException as req_err:
+        raise ValueError(f"An error occurred: {req_err}")
```
TheDZhon, line 48:
```suggestion
        raise ValueError(f"Request exception occurred: {req_err}")
```

diffyscan/utils/common.py
```diff
@@ -31,19 +33,38 @@ def load_config(path: str) -> Config:
...
+        response = requests.post(url, data=payload, headers=headers)
+        response.raise_for_status()
+    except requests.exceptions.HTTPError as http_err:
+        raise ValueError(f"HTTP error occurred: {http_err}")
+    except requests.exceptions.ConnectionError as conn_err:
+        raise ValueError(f"Connection error occurred: {conn_err}")
+    except requests.exceptions.Timeout as timeout_err:
+        raise ValueError(f"Timeout error occurred: {timeout_err}")
+    except requests.exceptions.RequestException as req_err:
+        raise ValueError(f"An error occurred: {req_err}")
```
TheDZhon, line 65:
```suggestion
        raise ValueError(f"Request exception occurred: {req_err}")
```

diffyscan/utils/common.py
```diff
@@ -57,3 +78,26 @@ def parse_repo_link(repo_link):
...
     return user_slash_repo
+
+
+def prettify_solidity(solidity_contract_content: str):
+    github_file_name = os.path.join(
+        tempfile.gettempdir(), "9B91E897-EA51-4FCC-8DAF-FCFF135A6963.sol"
+    )
+    with open(github_file_name, "w") as fp:
+        fp.write(solidity_contract_content)
+    prettier_return_code = subprocess.call(
```
TheDZhon, line 89:
```suggestion

    prettier_return_code = subprocess.call(
```

TheDZhon: to make it a bit easier for reading

diffyscan/utils/compiler.py
```diff
@@ -0,0 +1,111 @@
...
+from .helpers import create_dirs
+from .logger import logger
+
+
+def get_solc_native_platform_from_os():
+    platform_name = sys.platform
+    if platform_name == "linux":
+        return "linux-amd64"
+    elif platform_name == "darwin":
+        return "macosx-amd64" if platform.machine() == "x86_64" else "macosx-arm64"
```
TheDZhon, line 19:
```suggestion
    elif platform_name == "darwin":
        return "macosx-amd64" if platform.machine() == "x86_64" else "macosx-arm64"
    elif plaform_name == "win32":
        return "windows-amd64"
```

diffyscan/utils/compiler.py
```diff
@@ -0,0 +1,111 @@
...
+            compiler
+            for compiler in available_compilers_list["builds"]
+            if compiler["longVersion"] == required_compiler_version
+        ),
+        None,
+    )
+
+    if not required_build_info:
+        raise ValueError(
+            f'Required compiler version for "{required_platform}" is not found'
```
TheDZhon, line 38:
```suggestion
            f'Required compiler version "{required_compiler_version}" for "{required_platform}" is not found'
```

diffyscan/utils/compiler.py
```diff
@@ -0,0 +1,111 @@
...
+    try:
+        process = subprocess.run(
+            [compiler_path, "--standard-json"],
+            input=input_settings.encode(),
+            capture_output=True,
+            check=True,
+            timeout=30,
+        )
+    except subprocess.CalledProcessError as e:
+        raise ValueError(f"Error during subprocess execution: {e}")
```
TheDZhon, line 91:
```suggestion
        raise ValueError(f"Error during compiler subprocess execution: {e}")
```

diffyscan/utils/compiler.py
```diff
@@ -0,0 +1,111 @@
...
+    except subprocess.TimeoutExpired as e:
+        raise ValueError(f"Process timed out: {e}")
```
TheDZhon, line 93:
```suggestion
        raise ValueError(f"Compiler process timed out: {e}")
```

diffyscan/utils/constants.py
```diff
@@ -1,8 +1,52 @@
...
 START_TIME_INT = int(START_TIME)
 DIFFS_DIR = f"{DIGEST_DIR}/{START_TIME_INT}/diffs"
 LOGS_PATH = f"{DIGEST_DIR}/{START_TIME_INT}/logs.txt"
 DEFAULT_CONFIG_PATH = "config.json"
+
+SOLC_DIR = os.path.join(tempfile.gettempdir(), "solc_builds")
+
+GITHUB_API_TOKEN = os.getenv("GITHUB_API_TOKEN", "")
+if not GITHUB_API_TOKEN:
+    raise ValueError("GITHUB_API_TOKEN variable is not set")
```
TheDZhon, line 16: Sorry, but it doesn't look like a compile-time constant
I would reconsider placing this env var somewhere else

diffyscan/utils/encoder.py
```diff
@@ -0,0 +1,105 @@
...
+def encode_bytes32(data):
+    return data.replace("0x", "")
+
+
+def encode_bytes(data):
+    bytes_str = data.lstrip("0x")
+    data_length = len(bytes_str) // 2
+    encoded_length = 0
+    if data_length > 0:
+        encoded_length = ((len(bytes_str) - 1) // 64 + 1) * 64
```
TheDZhon, line 23: magic

diffyscan/utils/encoder.py
TheDZhon, line 1: I believe it's better to provide some documentation links here to understand the process of encoding

TheDZhon: This file is rock-hard for those who aren't familiar with the ABI encoder algorithms

diffyscan/utils/encoder.py
```diff
@@ -0,0 +1,105 @@
...
+                args_tuple_types = [
+                    component["type"]
+                    for component in constructor_abi[argument_index]["components"]
+                ]
+                if len(args_tuple_types) and args_tuple_types[0] == "address[]":
+                    dynamic_type_length = format(
+                        (len(constructor_calldata) // 64 + 1) * 32, "064x"
+                    )
+                    constructor_calldata += dynamic_type_length
+                    logger.info(f"dynamic_type_length {dynamic_type_length}")
```
TheDZhon, line 91: do we need this log?

diffyscan/utils/encoder.py
```diff
@@ -0,0 +1,105 @@
...
+                compl_data.append(length_arg)
+                compl_data.append(hex_text)
+            elif arg_type == "tuple":
+                args_tuple_types = [
+                    component["type"]
+                    for component in constructor_abi[argument_index]["components"]
+                ]
+                if len(args_tuple_types) and args_tuple_types[0] == "address[]":
+                    dynamic_type_length = format(
+                        (len(constructor_calldata) // 64 + 1) * 32, "064x"
```
TheDZhon, line 88: magic

diffyscan/utils/explorer.py
```diff
@@ -74,32 +93,85 @@ def _get_contract_from_mantle(mantle_explorer_hostname, contract):
...
+        result = _get_contract_from_mode(explorer_hostname, contract_address)
+    else:
+        result = _get_contract_from_etherscan(
+            token, explorer_hostname, contract_address
+        )
+
+    contract_name_from_etherscan = result["name"]
+    if contract_name_from_etherscan != contract_name_from_config:
+        raise ValueError(
+            f"Contract name in config does not match with Blockchain explorer {contract_address}: {contract_name_from_config} != {contract_name_from_etherscan}",
```
TheDZhon, line 146: something bad with formatting here :thinking:

diffyscan/utils/hardhat.py
```diff
@@ -0,0 +1,78 @@
...
+        pass
+
+    def is_port_in_use(self, parsed_url) -> bool:
+        import socket
+
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+            return s.connect_ex((parsed_url.hostname, parsed_url.port)) == 0
+
+    def start(self, main_config_relative_path: str, binary_config):
+        parsed_url = urlparse(binary_config["local_RPC_URL"])
```
TheDZhon, line 23: usually it's quite a bad idea to store RPC URL inside the config file because it might contain restricted access tokens that shouldn't be published in the sourcecode repo

the approach should be to load those:
- Etherscan token
- RPC URL(s)
- GitHub token

from private non-disclosed env :warning:

diffyscan/utils/hardhat.py
```diff
@@ -0,0 +1,78 @@
...
+            f"npx hardhat node --hostname {parsed_url.hostname} "
+            f"--port {parsed_url.port} "
+            f"--config {hardhat_config_relative_path} "
+            f"--fork {binary_config["remote_RPC_URL"]}"
+        )
+
+        logger.info(f'Trying to start Hardhat: "{local_node_command}"')
+        is_port_used = self.is_port_in_use(parsed_url)
+        if is_port_used:
+            answer = input(f'Port {parsed_url.port} is busy. Fix it? write "yes": ')
```
TheDZhon, line 39:
```suggestion
            answer = input(f'Port {parsed_url.port} is busy. Kill the app instance occupying the port? write "yes": ')
```

README.md
TheDZhon, line 1: The readme file should suggest the default config that supports binary bytecode comparison :pray:

TheDZhon (changes requested):
Thank you for taking the lead on implementing this feature #13 ! :+1:

The PR looks solid, and I appreciate the minimal use of external dependencies, which is crucial for avoiding supply chain vulnerabilities. :detective:

I’ve left a few suggestions and concerns in the comments that I believe would strengthen the implementation. Key points to consider:

* **Execution Flow**: We may need to think more deeply about the flow of execution, especially regarding the comparison process that involves checks without deployment and handling immutable logic, see my comment on this:

image


* **Edge Cases**: It would be helpful to account for some edge cases in the diffing process.
* **Documentation**: Please include the necessary documentation references for the ABI encoder.
* **Code Quality**: There’s some room for code improvement, which I’ve highlighted in the comments. :heart:

diffyscan/utils/binary_verifier.py
```diff
@@ -0,0 +1,142 @@
...
+    differences = []
+    for i in range(maxInstructionsCount):
+        actual = actualInstructions[i] if i < len(actualInstructions) else None
+        expected = expectedInstructions[i] if i < len(expectedInstructions) else None
+        if not actual and not expected:
+            raise ValueError("Invalid instructions data")
+        elif (actual is not None) and (
+            actual.get("bytecode") != expected.get("bytecode")
+        ):
+            differences.append(i)
```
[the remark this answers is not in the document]
tamtamchik: The same is true for `actual is None` and some `expected` should also be added to the differences.

diffyscan/utils/binary_verifier.py
```diff
@@ -0,0 +1,142 @@
...
+        else:
+            raise ValueError("Invalid bytecode difference data")
+
+
+def parse(bytecode):
+    buffer = bytes.fromhex(bytecode[2:] if bytecode.startswith("0x") else bytecode)
+    instructions = []
+    i = 0
+    while i < len(buffer):
+        opcode = buffer[i]
```
tamtamchik, line 131: Consider adding an edge case for not valid opcode (e.g., `if opcode not in OPCODES`)

diffyscan/utils/binary_verifier.py
tamtamchik: Plus, it is worth adding a check that the actual length is enough for PUSH(X), wdyt?

diffyscan/utils/binary_verifier.py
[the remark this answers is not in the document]
TheDZhon: Might be solve with the following chunk (warn: untested)

```python
    if actual != expected:
        if actual is None or expected is None:
            differences.append(i)
        elif actual.get("bytecode") != expected.get("bytecode"):
            differences.append(i)
```

config_samples/lido_dao_holesky_config.json
```diff
@@ -111,5 +112,41 @@
             "commit": "6bd6b76d1156e20e45d1016f355d154141c7e5b9",
             "relative_root": "contracts"
         }
+    },
+    "binary_checking": {
+	"enable": true,
```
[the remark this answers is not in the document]
Alexvozhak: done

config_samples/lido_dao_holesky_config.json
```diff
@@ -111,5 +112,41 @@
...
+	"skip_deploy_error" : false,
```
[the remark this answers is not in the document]
Alexvozhak: if we want to find all binary comparison errors for all contracts in one pass

config_samples/lido_dao_holesky_config.json
```diff
@@ -111,5 +112,41 @@
...
+	"hardhat_config_name": "holesky_hardhat.config.js",
```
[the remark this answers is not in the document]
Alexvozhak: do you mean the [same](https://github.com/lidofinance/diffyscan/pull/52#discussion_r1705345906) point?

diffyscan/diffyscan.py
```diff
@@ -3,39 +3,121 @@
...
+        config["binary_checking"],
+    )
+    if calldata is not None:
+        contract_creation_code += calldata
+    elif text_error is not None:
+        skip_or_raise(skip_deploy_error, text_error)
+        return
+
+    local_RPC_URL = config["binary_checking"]["local_RPC_URL"]
+    local_contract_address, text_reason = deploy_contract(
```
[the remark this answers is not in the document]
Alexvozhak: done

diffyscan/diffyscan.py
```diff
@@ -149,28 +230,90 @@ def process_config(path: str, recursive_parsing: bool, unify_formatting: bool):
...
+            config["explorer_token_env_var"], masked=True, required=False
+        )
+    if explorer_token is None:
+        logger.warn(
+            f'Failed to find explorer token in config ("explorer_token_env_var")'
+        )
+        explorer_token = os.getenv("ETHERSCAN_EXPLORER_TOKEN", default=None)
+    if explorer_token is None:
+        raise ValueError(
+            f'Failed to find explorer token in env ("ETHERSCAN_EXPLORER_TOKEN")'
```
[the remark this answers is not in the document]
Alexvozhak: done

diffyscan/diffyscan.py
```diff
@@ -149,28 +230,90 @@ def process_config(path: str, recursive_parsing: bool, unify_formatting: bool):
...
+                )
+    except KeyboardInterrupt:
+        logger.info(f"Keyboard interrupt by user")
+    finally:
+        if binary_check:
+            hardhat.stop()
 
 
 def parse_arguments():
     parser = argparse.ArgumentParser()
```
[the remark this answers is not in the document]
Alexvozhak: done

diffyscan/diffyscan.py
Alexvozhak: done

diffyscan/utils/calldata.py
Alexvozhak: reduced to: "Constructor in ABI successfully found"

diffyscan/utils/calldata.py
Alexvozhak: done

diffyscan/utils/calldata.py
Alexvozhak: reworked with  exceptions usage

diffyscan/utils/constants.py
Alexvozhak: done

diffyscan/utils/encoder.py
Alexvozhak: removed

diffyscan/utils/explorer.py
Alexvozhak: done

diffyscan/utils/hardhat.py
Alexvozhak: done

diffyscan/diffyscan.py
```diff
@@ -3,39 +3,121 @@
...
 import os
-import subprocess
-import tempfile
 
-from .utils.common import load_config, load_env
-from .utils.constants import DIFFS_DIR, START_TIME, DEFAULT_CONFIG_PATH
-from .utils.explorer import get_contract_from_explorer
-from .utils.github import get_file_from_github, get_file_from_github_recursive, resolve_dep
+from .utils.common import load_config, load_env, prettify_solidity
+from .utils.constants import *
```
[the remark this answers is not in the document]
Alexvozhak: done

Alexvozhak:
@TheDZhon fixed all remarks except encoder.py and binary_verifier.py. I will fix it for a few days

diffyscan/utils/binary_verifier.py
```diff
@@ -0,0 +1,142 @@
+from .logger import *
+from .constants import OPCODES
+
+
+def to_match(actualBytecode, expectedBytecode, immutables):
+    logger.info("Comparing actual code with the expected one...")
+
+    actualInstructions = parse(actualBytecode)
+    expectedInstructions = parse(expectedBytecode)
+    maxInstructionsCount = max(len(actualInstructions), len(expectedInstructions))
```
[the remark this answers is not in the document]
Alexvozhak: done

diffyscan/utils/binary_verifier.py
```diff
@@ -0,0 +1,142 @@
...
+
+    if actualInstructions:
+        checkpoints.add(len(actualInstructions) - 1)
+
+    if expectedInstructions:
+        checkpoints.add(len(expectedInstructions) - 1)
+
+    for ind in list(checkpoints):
+        startIndex = max(0, ind - nearLinesCount)
+        lastIndex = min(ind + nearLinesCount, maxInstructionsCount - 1)
```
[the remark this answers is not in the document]
Alexvozhak: done

diffyscan/utils/binary_verifier.py
```diff
@@ -0,0 +1,142 @@
...
+        for i in range(startIndex, lastIndex + 1):
+            checkpoints.add(i)
```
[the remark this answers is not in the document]
Alexvozhak: done

diffyscan/utils/binary_verifier.py
Alexvozhak: done

diffyscan/utils/encoder.py
Alexvozhak: source of magic [here](https://github.com/lidofinance/lido-dao/blob/a9a9ea50b5be7dba33e06d70cc068252ffd01b52/bytecode-verificator/bytecode_verificator.sh#L239)

Alexvozhak: now it seems more readable

diffyscan/utils/encoder.py
Alexvozhak: tried to do more readable

diffyscan/utils/encoder.py
Alexvozhak: added link into comments, also i am going to add it into the readme

README.md
Alexvozhak: done!

README.md
```diff
@@ -35,6 +41,17 @@ Set your Github token to query API without strict rate limiting,
 export GITHUB_API_TOKEN=<your-github-token>
 ```
 
+Set remote RPC URL to validate contract bytecode at remote rpc node,
+
+```bash
+export REMOTE_RPC_URL =<remote-rpc-url>
+```
+
+Set local RPC URL to validate contract bytecode at local rpc node (Hardhat),
```
TheDZhon, line 50: This description is misleading, we need to have LOCAL_RPC_URL mostly to address the case when immutables should be checked against the local deployment and provided ctor arguments, isn't it?

README.md
```diff
@@ -62,9 +79,54 @@ Alternatively, create a new config file named `config.json`,
             "commit": "6bd6b76d1156e20e45d1016f355d154141c7e5b9",
             "relative_root": "contracts"
         }
+    },
+    "raise_exception": true,
```
TheDZhon, line 83: What does 'raise_exception' mean in this context, appearing out of the blue?

README.md
```diff
@@ -62,9 +79,54 @@ Alternatively, create a new config file named `config.json`,
...
+    "bytecode_comparison": {
+        "hardhat_config_name": "holesky_hardhat.config.js",
+        "constructor_calldata": {
+            "0x28FAB2059C713A7F9D8c86Db49f9bb0e96Af1ef8": "000000000000000000000000ab89ed3d8f31bcf8bb7de53f02084d1e6f043d34000000000000000000000000e92329ec7ddb11d25e25b3c21eebf11f15eb325d00000000000000000000000000000000000000000000000000000000000000600000000000000000000000000000000000000000000000000000000000000000"
+        },
+        "constructor_args": {
+            "0x28FAB2059C713A7F9D8c86Db49f9bb0e96Af1ef8": [
+                "0xab89ED3D8f31bcF8BB7De53F02084d1e6F043D34",
+                "0xE92329EC7ddB11D25e25b3c21eeBf11f15eB325d",
+                ""
```
TheDZhon, line 93: Is it a good example when we have both args and calldata for the same contract address? If so, what would be the order of preference?

In my view, the program should treat this config being ambiguous and invalid (malformed) :warning:

README.md
```diff
@@ -62,9 +79,54 @@ Alternatively, create a new config file named `config.json`,
...
+                    "0xc7cc160b58F8Bb0baC94b80847E2CF2800565C50",
+                    "0xF0179dEC45a37423EAD4FaD5fCb136197872EAd9",
+                    "0xC01fC1F2787687Bc656EAc0356ba9Db6e6b7afb7"
+                ]
+            ]
+        }
     }
 }
 ```
+use `holesky_hardhat.config.js` from `config.json` or create near the diffyscan.py own Hardhat config `hardhat_config.js`:
```
TheDZhon, line 117: Suppose I installed diffyscan using the command from the Install section of README:
```
pipx install git+https://github.com/lidofinance/diffyscan
```

Where should I place this, then?

config_samples/lido_dao_holesky_config.json
```diff
@@ -1,115 +1,175 @@
...
+      "commit": "fa64a1ced0b70ab89073d5d0b6e01b0778f7e7d6",
+      "relative_root": "contracts"
+    },
+    "@openzeppelin/contracts-v4.4": {
+      "url": "https://github.com/OpenZeppelin/openzeppelin-contracts",
+      "commit": "6bd6b76d1156e20e45d1016f355d154141c7e5b9",
+      "relative_root": "contracts"
+    }
+  },
+  "raise_exception": true,
```
TheDZhon, line 116: I find this name being too vague tbh (as said above)

config_samples/lido_dao_holesky_config.json
```diff
@@ -1,115 +1,175 @@
...
+    "hardhat_config_name": "holesky_hardhat_config.js",
+    "constructor_calldata": {
+      "0x28FAB2059C713A7F9D8c86Db49f9bb0e96Af1ef8": "000000000000000000000000ab89ed3d8f31bcf8bb7de53f02084d1e6f043d34000000000000000000000000e92329ec7ddb11d25e25b3c21eebf11f15eb325d00000000000000000000000000000000000000000000000000000000000000600000000000000000000000000000000000000000000000000000000000000000"
+    },
+    "constructor_args": {
+      "0x28FAB2059C713A7F9D8c86Db49f9bb0e96Af1ef8": [
+        "0xab89ED3D8f31bcF8BB7De53F02084d1e6F043D34",
+        "0xE92329EC7ddB11D25e25b3c21eeBf11f15eB325d",
+        ""
+      ],
```
TheDZhon, line 127: once again, it's a duplicate address

config_samples/lido_dao_sepolia_config.json
```diff
@@ -1,101 +1,298 @@
...
+  "bytecode_comparison": {
+    "hardhat_config_name": "sepolia_hardhat_config.js",
+    "constructor_calldata": {
+      "0x8f6254332f69557A72b0DA2D5F0Bc07d4CA991E7": "000000000000000000000000f30a674935479cc6f2254ba65d7534eab8bd6ba20000000000000000000000006885e36bfcb68cb383dfe90023a462c03bcb2ae500000000000000000000000000000000000000000000000000000000000000600000000000000000000000000000000000000000000000000000000000000000"
+    },
+    "constructor_args": {
+      "0x8f6254332f69557A72b0DA2D5F0Bc07d4CA991E7": [
+        "0xf30a674935479cc6f2254ba65d7534eab8bd6ba2",
+        "0x6885E36BFcb68CB383DfE90023a462C03BCB2AE5",
+        ""
```
TheDZhon, line 111: duplicate :thinking:

diffyscan/diffyscan.py
```diff
@@ -3,39 +3,106 @@
...
+        remote_deployed_bytecode,
+        immutables,
+    )
+
+    if is_fully_matched:
+        return
+
+    logger.info(
+        f"Trying to check bytecode via using calldata and deploying into local node"
+    )
```
TheDZhon, line 71: Shall say here that automated match hasn't worked out

diffyscan/diffyscan.py
```diff
@@ -145,32 +211,127 @@ def run_diff(config, name, address, explorer_api_token, github_api_token, recurs
...
+
+    github_api_token = os.getenv("GITHUB_API_TOKEN", "")
+    if not github_api_token:
+        raise ValueError("GITHUB_API_TOKEN variable is not set")
+
+    local_rpc_url = os.getenv("LOCAL_RPC_URL", "")
+    if not local_rpc_url:
+        raise ValueError("LOCAL_RPC_URL variable is not set")
+    if not local_rpc_url.startswith("http://"):
+        local_rpc_url = "http://" + local_rpc_url
```
TheDZhon, line 250: why do we do this :) It's a bit surprising and inconsistent when looking down at the `REPOTE_RPC_URL` case, e.g.

diffyscan/utils/common.py
```diff
@@ -57,3 +79,27 @@ def parse_repo_link(repo_link):
     repo_location = [item.strip("/") for item in parse_result[2].split("tree")]
     user_slash_repo = repo_location[0]
     return user_slash_repo
+
+
+def prettify_solidity(solidity_contract_content: str):
+    github_file_name = os.path.join(
+        tempfile.gettempdir(), "9B91E897-EA51-4FCC-8DAF-FCFF135A6963.sol"
```
TheDZhon, line 86: Can't we generate a random uuid here for the filename?

diffyscan/utils/encoder.py
```diff
@@ -0,0 +1,121 @@
...
+def encode_bytes32(data: str) -> str:
+    return data.replace("0x", "")
+
+
+def encode_bytes(data: str) -> str:
+    bytes_str = data.lstrip("0x")
+    if not bytes_str:
+        return to_hex_with_alignment(0)
+
+    # untested, taken from here:
```
TheDZhon, line 22: huh?

diffyscan/utils/encoder.py
```diff
@@ -0,0 +1,121 @@
...
+    encoded_value_length = to_hex_with_alignment(len(encoded_value))
+    return (
+        offset_to_start_of_data_part,
+        encoded_value_length,
+        encoded_value.hex().ljust(64, "0"),
+    )
+
+
+def encode_constructor_arguments(constructor_abi: list, constructor_config_args: list):
+    # see https://docs.soliditylang.org/en/develop/abi-spec.html#contract-abi-specification
```
TheDZhon, line 75: it's good to have this comment though might have been a bit more specific

diffyscan/utils/encoder.py
TheDZhon: Can it be commented with a brief explanation of the essence of this operation instead of saying 'taken from there and untested'? :pray:

diffyscan/utils/binary_verifier.py
TheDZhon: I'd say it should be a warning only

README.md
Alexvozhak: renamed to raise_exceptions_on_comparison

README.md
Alexvozhak: i've prefered prepared calldata from field 'constructor_calldata' to raw calldata in 'constructor_args'. Fixed example

README.md
Alexvozhak: the readme has been updated in this part

config_samples/lido_dao_holesky_config.json
Alexvozhak: raise_exceptions_on_comparison

config_samples/lido_dao_holesky_config.json
Alexvozhak: the duplicate from 'constructor_args' has been removed

config_samples/lido_dao_sepolia_config.json
Alexvozhak: fixed

diffyscan/diffyscan.py
Alexvozhak: fixed

diffyscan/diffyscan.py
Alexvozhak: done

diffyscan/utils/common.py
Alexvozhak: done

README.md
Alexvozhak: Set local RPC URL to check immutables against the local deployment and provided constructor arguments

diffyscan/utils/encoder.py
Alexvozhak: added [link ](https://github.com/lidofinance/lido-dao/blob/master/bytecode-verificator/bytecode_verificator.sh#L369-L405)

diffyscan/utils/encoder.py
Alexvozhak: i'm waiting the example of specific contract

README.md
TheDZhon: 1) It's still not clear — maybe it's better to describe the outcome, not the mechanism?
e.g.: `fail_on_comparison_error`
2) somewhere it's renamed, somewhere (like here) it's not :thinking:

README.md
TheDZhon: not sure if it's a good approach still: it's too ambiguous and undocumented, why we can't enforce a uniqueness?

README.md
```diff
@@ -62,9 +80,54 @@ Alternatively, create a new config file named `config.json`,
...
+                    "0xc7cc160b58F8Bb0baC94b80847E2CF2800565C50",
+                    "0xF0179dEC45a37423EAD4FaD5fCb136197872EAd9",
+                    "0xC01fC1F2787687Bc656EAc0356ba9Db6e6b7afb7"
+                ]
+            ]
+        }
     }
 }
 ```
+then create a new Hardhat config file named `hardhat_config.js` near the diffyscan.py
```
TheDZhon, line 118: What if you installed diffyscan via pipx as suggested in README? Where should I place this file?

diffyscan/utils/binary_verifier.py
```diff
@@ -0,0 +1,142 @@
...
+            actual.get("bytecode") != expected.get("bytecode")
+        ):
+            differences.append(i)
+
+    if not differences:
+        logger.okay(f"Bytecodes are fully matched")
+        return
+    logger.warn(f"Bytecodes have differences")
+
+    nearLinesCount = 3
```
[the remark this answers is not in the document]
TheDZhon: doubling down

diffyscan/utils/encoder.py
TheDZhon: here: https://sepolia-optimism.etherscan.io/address/0xf49D208B5C7b10415C7BeAFe9e656F2DF9eDfe3B

init tx: https://sepolia-optimism.etherscan.io/tx/0x4c9db6f7c0e2ca4b831bfc67f46f265a1bdac5735738900e42014c88ef361373

diffyscan/utils/encoder.py
TheDZhon:
```python
# Calculate the length of the hex-encoded 32-bytes padded data 
# since EVM uses 32-byte (256-bit) words
```

TheDZhon (changes requested):
:eyes:

README.md
```diff
@@ -62,9 +80,54 @@ Alternatively, create a new config file named `config.json`,
...
+```json
+module.exports = {
+  solidity: "0.8.9",
+  networks: {
+    hardhat: {
+      chainId: 17000,
+      blockGasLimit: 92000000,
+      hardfork: "istanbul",
```
TheDZhon, line 126:
```suggestion
      hardfork: "cancun",
```

TheDZhon: not critical though

package.json
```diff
@@ -1,6 +1,16 @@
 {
```
TheDZhon, line 1: maybe add `husky` for an automated setup of a precommit hook

diffyscan/utils/encoder.py
TheDZhon: https://docs.optimism.io/chain/networks#op-sepolia

TheDZhon (commented):
:eyes:

README.md
Alexvozhak: done

diffyscan/utils/binary_verifier.py
```diff
@@ -0,0 +1,159 @@
...
+    actual_instructions, unknown_opcodes_first_half = parse(actual_bytecode)
+    expected_instructions, unknown_opcodes_second_half = parse(expected_bytecode)
+
+    unknown_opcodes = (
+        unknown_opcodes_first_half or set() | unknown_opcodes_second_half or set()
+    )
+    if unknown_opcodes:
+        logger.warn(f"Detected unknown opcodes: {unknown_opcodes}")
+
+    zipped_instructions = list(zip(actual_instructions, expected_instructions))
```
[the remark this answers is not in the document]
Alexvozhak: done

README.md
Alexvozhak: done

package.json
Alexvozhak: done

TheDZhon:
Seems the PR is ready to merge :star2: except only for two things:
1) An outstanding branch conflict it has
2) Metadata parsed as opcodes while it shouldn't, see https://docs.soliditylang.org/en/latest/metadata.html#encoding-of-the-metadata-hash-in-the-bytecode
image 

see as a reference:

> https://playground.sourcify.dev/
> address: 0x76A50b8c7349cCDDb7578c6627e79b5d99D24138
> network: #10 OP Mainnet
> image
>

TheDZhon:
https://github.com/lidofinance/aave-delivery-infrastructure/blob/main/verifications/scripts/verify-bytecode.ts#L479-L494
here is the example of the metadata parsing btw

TheDZhon (approved):
LGTM, accepting to upstream :green_apple:

PR #60, opened 2024-08-17
Cfg: add new mainnet addresses for stETH on Optimism upgrade

kovalgek:
Add new mainnet addresses for stETH on Optimism upgrade.


TheDZhon (approved):
LGTM

PR #62, opened 2024-10-08
feat: add sr2 verify scripts

Amuhar:


config_samples/sr-2/mainnet/core_sr_v2_mainnet_config.json
```diff
@@ -0,0 +1,72 @@
+{
+  "contracts": {
```
TheDZhon, line 2: I guess that MinFirstAllocationStrategy is missing here

TheDZhon (changes requested):
:eyes:

PR #63, opened 2024-10-15
Develop to Main

TheDZhon:
based on #52 with the following changes:
- opt-in
- mask RPC URLs
- a few chores for error codes


hardhat_configs/holesky_hardhat_config.js
tamtamchik, line 1: I would like to propose having a comment somewhere on why we stick to files (AFAIR, it was a bug in hardhat, so it might be fixed, github issue will do).

diffyscan/utils/constants.py
```diff
@@ -1,8 +1,62 @@
...
+# fmt: on
+
+
+def get_key_from_value(dictinary: dict, value: str):
+    keys = [
+        dict_key for dict_key, dict_value in dictinary.items() if dict_value == value
+    ]
+    if keys:
+        return keys[0]
+    return None
```
tamtamchik, line 58: Not sure about it, but its more readable, imo

```suggestion
def get_key_from_value(dictionary: dict, value: str):
    return next((k for k, v in dictionary.items() if v == value), None)
```

tamtamchik (approved):
Looks good! No major issues.

Just small suggestions.

.husky/pre-commit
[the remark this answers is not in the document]
TheDZhon: Fixed :+1:

diffyscan/diffyscan.py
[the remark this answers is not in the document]
TheDZhon: :eyes: what a keen eye

hardhat_configs/holesky_hardhat_config.js
TheDZhon: done :+1:

diffyscan/utils/hardhat.py
```diff
@@ -0,0 +1,96 @@
...
+        except subprocess.TimeoutExpired:
+            self._handle_timeout(parsed_url)
+
+    def stop(self):
+        if self.sub_process is not None and self.sub_process.poll() is None:
+            os.kill(self.sub_process.pid, signal.SIGTERM)
+            logger.info(f"Hardhat stopped, PID {self.sub_process.pid}")
+
+    def _is_port_in_use_(self, parsed_url) -> bool:
+        import socket
```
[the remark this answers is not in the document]
TheDZhon: good catch!

diffyscan/utils/constants.py
TheDZhon: Yead, fixed!

PR #74, opened 2025-01-23
Cfg: add wstETH on Lisk

ricott1:
- [x] add config files for Lisk Sepolia
- [x] add config files for Lisk Mainnet


TheDZhon (approved):
LGTM

PR #75, opened 2025-02-05
Fix: add CODEOWNERS

TheDZhon:


tamtamchik (approved):
LGTM ✅ 

We do not have automation in this repo, but for future it is OK.

PR #87, opened 2025-05-14
Feat/easy track

Alexvozhak:
While working on EasyTrack contract support, **three** bugs were detected and fixed:

1. Unnecessary, unexpected slicing caused by a typo (contract["solcInput":] instead of contract["solcInput"]), which led to malformed input objects (see [here](https://github.com/lidofinance/diffyscan/commit/98a1f5169167c020c2cff0329c16521a4bd7e2c7))
2. Encoding error when attempting to encode a string as a BigInt in a tuple, resulting in runtime failures (see [here](https://github.com/lidofinance/diffyscan/commit/1fd8c59755777bd0063c7b18365b2bb600eafc3e)). 

Verification of the following contracts from Sepolia failed with an error:
```
0x61Bb0Ef69262d5EF1cc2873cf61766751D99B699
0x082d16150BF75BB8F2197eEC1d293DbA96c93638
0x758D8c3CE794b3Dfe3b3A3482B7eD33de2109D95
0xC40801b88C835a58e54eEE6679D301ba31a4C72b
0x098a952BD200005382aEb3229e38ae39A7616F56
0xbac2A471443F18aC5C31078b96C5797A78fCc680
```
image_2025-05-14_12-46-11

3. Missing calldata in config was an error if constructor has no arguments in ABI (see [here](https://github.com/lidofinance/diffyscan/pull/87/commits/a5163e73385006f6148b802f3b027f1f5bd6155e))

One feature was added also:
The "source_comparison" flag was added to config, which disables source comparison (e.g. EasyTrack or other monolithic contracts), see [here](https://github.com/lidofinance/diffyscan/pull/87/commits/5149651e3d5aefcfa1f42f99a5706674a8b4b4e0)

A follow-up commit made for transparency, in which the configuration flag fail_on_comparison_error was renamed to fail_on_bytecode_comparison_error, see [here](https://github.com/lidofinance/diffyscan/pull/87/commits/4d7668d940c82539616dd761ab3fccc11833bbd3)


diffyscan/diffyscan.py
```diff
@@ -106,6 +106,18 @@ def run_source_diff(
     recursive_parsing=False,
     prettify=False,
 ):
+    source_files = (
+        contract_code["solcInput"].items()
+        if "sources" not in contract_code["solcInput"]
+        else contract_code["solcInput"]["sources"].items()
+    )
+
+    if is_monolithic_contract(source_files):
```
TheDZhon, line 115: I'd suggest to shape the func checking if the contract machine-friendly to verify, see: https://docs.soliditylang.org/en/latest/using-the-compiler.html#compiler-input-and-output-json-description 

```suggestion
    if not is_standard_json_contract(source_files):
```

diffyscan/diffyscan.py
```diff
@@ -386,5 +391,34 @@ def main():
     logger.okay(f"Done in {round(execution_time, 3)}s ✨" + " " * 100)
 
 
+def is_monolithic_contract(source_files: dict) -> bool:
```
TheDZhon, line 394:
```suggestion
def is_standard_json_contract(source_files: dict) -> bool:
```

diffyscan/diffyscan.py
```diff
@@ -106,6 +106,18 @@ def run_source_diff(
...
 ):
+    source_files = (
+        contract_code["solcInput"].items()
+        if "sources" not in contract_code["solcInput"]
+        else contract_code["solcInput"]["sources"].items()
+    )
+
+    if is_monolithic_contract(source_files):
+        logger.warn(
+            f"Contract {contract_code['name']} is monolithic, skipping source code diff"
```
TheDZhon, line 117:
```suggestion
            f"Contract {contract_code['name']} isn't verified with a standard json format, skipping source code diff"
```

diffyscan/diffyscan.py
```diff
@@ -386,5 +391,34 @@ def main():
     logger.okay(f"Done in {round(execution_time, 3)}s ✨" + " " * 100)
 
 
+def is_monolithic_contract(source_files: dict) -> bool:
+    """
+    Determines if the contract is monolithic (single-file/flattened mode).
```
TheDZhon, line 396: can we try to retrieve the format information from the Etherscan API?

TheDZhon (changes requested):
Great work, let's polish it a bit!

diffyscan/diffyscan.py
Alexvozhak: I have just researched this problem - no, we can't retrieve it from the Etherscan API

config_samples/ethereum/hoodi/hoodi_easy_track_config.json
```diff
@@ -0,0 +1,31 @@
...
+  "explorer_token_env_var": "ETHERSCAN_TOKEN",
+  "github_repo": {
+    "url": "https://github.com/lidofinance/easy-track",
+    "commit": "564700faaa6a8af5bfac24314414ca99244a069f",
+    "relative_root": "contracts"
+  },
+  "dependencies": {},
+  "fail_on_comparison_error": true,
+  "source_comparison": false,
+  "bytecode_comparison": {
```
TheDZhon, line 16: 1) what would happen if I set this:

```
  "fail_on_comparison_error": false,
  "source_comparison": false,
  "bytecode_comparison": {
```

2) does this flag `fail_on_comparison_error` work only for bytecode comparison or for source too?

diffyscan/diffyscan.py
```diff
@@ -255,6 +258,12 @@ def process_config(
 
         ExceptionHandler.initialize(config["fail_on_comparison_error"])
 
+    enable_source_comparison = config.get("source_comparison", True)
+    if not enable_source_comparison:
+        logger.warn(
+            f'Source code comparison is disabled in {path}. To enable, set "source_comparison": true in your config'
```
TheDZhon, line 264:
```suggestion
            f'Source code comparison is disabled in {path}. To enable, set "source_comparison": true in the config'
```

TheDZhon (changes requested):
Good, just a single question about config flags and their mutual effects and corner cases.

config_samples/ethereum/hoodi/hoodi_easy_track_config.json
Alexvozhak: With "fail_on_comparison_error": false, you’re telling DiffyScan not to suppress errors in either bytecode or source comparisons—so any exception will immediately be raised and abort the scan (see [here](https://github.com/lidofinance/diffyscan/blob/2eb79307b36fc1e5522c417c9cac1148f3bede17/diffyscan/diffyscan.py#L258-L295))

Alexvozhak: There’s no point in applying the "fail_on_comparison_error" flag to source code checks. The custom exceptions affected by this flag are only raised in a single place (see [here](https://github.com/lidofinance/diffyscan/blob/2eb79307b36fc1e5522c417c9cac1148f3bede17/diffyscan/utils/common.py#L38-L52))

These exceptions are handled based on the flag, but the scenario where the explorer fails on one contract (an exception is raised) and then successfully processes the next one is highly unlikely.

After discussing with @TheDZhon, we decided to rename the fail_on_comparison_error flag. 
I’m proposing to rename it to fail_on_bytecode_comparison_error (commit [here](https://github.com/lidofinance/diffyscan/pull/87/commits/4d7668d940c82539616dd761ab3fccc11833bbd3)).

TheDZhon (approved):
LGTM

PR #165, opened 2026-06-11
chore: add dependabot config

tamtamchik:
Adds a Dependabot configuration for weekly grouped dependency updates:

- `github-actions`: weekly, grouped, 7-day cooldown
- `uv`: weekly, all dependencies grouped, 7-day cooldown

Workflow actions are already pinned to commit hashes.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,28 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    cooldown:
+      default-days: 7
+    groups:
+      github-actions:
+        patterns: ["*"]
+    labels: ["dependencies", "github-actions"]
+    commit-message:
+      prefix: "ci"
+
+  - package-ecosystem: "uv"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    cooldown:
+      default-days: 7
+    groups:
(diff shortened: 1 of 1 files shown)


TheDZhon (approved):
LGTM

TheDZhon:
A few non-blocking follow-ups worth considering (none should hold up the merge):

1. Add a docker ecosystem block for the devcontainer base image. .devcontainer/Dockerfile pins the base image by digest (mcr.microsoft.com/devcontainers/python:3.11@sha256:…), and that image is built and run in regression.yml — so it's effectively part of our trusted execution environment for verification. The current config won't keep it updated; a docker block (directory /.devcontainer) would, the snippet could be like:
```
   - package-ecosystem: "docker"
    directory: "/.devcontainer"
    schedule:
      interval: "weekly"
    cooldown:
      default-days: 7
    labels: ["dependencies", "docker"]
    commit-message:
      prefix: "build"
```
2. File a note about the manually-pinned UV_VERSION drift. UV_VERSION=0.10.9 is hardcoded in both .devcontainer/Dockerfile and regression.yml. The uv ecosystem here bumps uv.lock/pyproject.toml deps but won't touch the uv binary version, so the resolver can drift from the locked deps it produced. Out of scope for this PR — just worth tracking in an issue so "Dependabot keeps us current" doesn't give a false sense of completeness.

PR #188, opened 2026-07-06
chore: add hyper native guard config

dry914:
Add config for Hypernative guard integration


tamtamchik (dismissed):
LGTM, only consider moving to yaml

tamtamchik (approved):

PR #189, opened 2026-07-13
chore: add SM+SR ETs

madlabman:
Covers CSMv3, CMv2 and SR ET factories

PR #191, opened 2026-07-13
chore: add SM+SR ET factories config

tamtamchik:
Builds on #189: converts the config to YAML, pins the audited commit with a link to the [07-2026 MixBytes Easy Track Factories audit](https://github.com/lidofinance/audits#07-2026-mixbytes-lido-easy-track-factories-srv3-csmv3-cmv2-security-audit), tags the OZ dependency as v4.3.2, and declares the known diffs:

- source: verified sources on Etherscan carry root-relative imports, GitHub uses relative ones (import hunks only, scoped via `line_ranges`);
- bytecode: `cbor_metadata` only — the metadata hash covers source file hashes, so the rewritten imports change it while the runtime code matches byte-for-byte.

Run is green: 9 allowed source + 9 allowed bytecode diffs, 0 failures.


TheDZhon (approved):
LGTM, should it be in CI or not though?

PR #192, opened 2026-07-16
feat: add local_compilation mode for explorer-unverified contracts

arwer13:
Contracts **without explorer-verified** source can now **be checked at the bytecode level**: local_compilation.inputs maps an address to a local solc standard-JSON input (source manifest + compiler settings), and local_compilation.compiler pins the solc build. Source contents are still fetched from the pinned GitHub commit, and the source diff is skipped for these addresses since there is nothing explorer-side to diff against.

## Relation to `feat/local-compilation-mode`

This PR implements the same feature as in `feat/local-compilation-mode` and **keeps its config
schema verbatim** (`local_compilation: {compiler, inputs}`), so any config written
for either branch works with both. We propose landing this implementation because:

1. **Rebased on current `main`.** The earlier branch forked ~82 commits back,
   before the `allowed_diffs` machinery and related `process_config` refactors,
   and no longer applies cleanly. This one is written against today's `main` and
   composes with allowlists, caching, and the summary/suggestion output.

2. **Local contracts reuse the explorer normalization path.** The payload is built
   via the existing `_build_contract_payload`/`_attach_contract_metadata` helpers
   instead of by hand, so EVM-version normalization and library re-keying to the
   definition file (required for correct solc linking) apply to local inputs
   exactly as they do to explorer bundles.

3. **More guardrails.** All-local configs need no explorer hostname/chain-id/token
   at all; unknown addresses in `local_compilation.inputs` are flagged; a config
   whose local contracts would get zero verification (bytecode comparison off)
   warns; the skipped source diff is logged per contract rather than silent;
   malformed/empty solc inputs are rejected with clear errors.

4. **Docs and tests included.** README section (schema + why a full-bytecode match
   incl. the CBOR metadata tail is sound evidence for guessed settings), CLAUDE.md
   schema note, and tests covering error paths (malformed JSON, empty sources,
   absolute/relative input paths).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -69,6 +69,9 @@ explorer_token_env_var: "ETHERSCAN_EXPLORER_TOKEN"
 github_repo:        { url, commit, relative_root }
 dependencies:       { "dep_name": { url, commit, relative_root } }
 bytecode_comparison: { constructor_calldata, constructor_args, libraries, deployment_from, extra_sources }
+local_compilation:  { compiler, inputs }  # optional; bytecode-only checks for explorer-unverified contracts
+                    # compiler: full solc build string ("v0.8.20+commit.a1b79de6"); inputs: { "0xaddr": "path/to/solc-input.json" } relative to the config file
+                    # the input supplies the source manifest + settings; contents still come from GitHub; source diff is skipped for these addresses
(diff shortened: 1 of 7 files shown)


tamtamchik (changes requested):
The issue is that if you have sources to create an identical bytecode, you can put it to etherscan / blockscout / sourcify as verified. So I don't think this PR is actually worth to be included.

TheDZhon:
Closing as the use case seems to be too exotic.

PR #205, opened 2026-08-26
feat: add Robinhood earnUSDe deployment config

tamtamchik:
## Summary

- add Robinhood Chain core and earnUSDe configs
- pin the [Nethermind-audited flexible-vaults commit](https://github.com/lidofinance/audits/blob/main/earn/2026-03-02_Nethermind_NM-0758_Mellow.pdf) and OpenZeppelin v5.3.0
- allow expected per-deployment ProxyAdmin immutables
- add ROBINHOOD_RPC_URL to .env.example
- allow known live-state drift in the V3 voting and SR v3 regression configs


TheDZhon:
Codex:

1. **An additional EOA can replace the verifier permission root directly.**

   The [vault initializer](https://github.com/lidofinance/diffyscan/blob/50113b1a2851ad34488ee3710383f3dfafd2fe8e/config_samples/robinhood/mainnet/meta/meta_usde_robinhood_config.yaml#L103) grants `SET_MERKLE_ROOT_ROLE` both to the timelock and to `0xE98B…95A3`. At finalized Robinhood block `47,542,770` (`0xa398280eb18fe6bb8790d08a8c441ce5a0e4d4dab1a73dbd3a6e8808e2449c94`), two independent RPC providers returned the role as active for both holders, and `0xE98B…95A3` had no deployed bytecode. The address is not part of the expected actor inventory.

   [`Verifier.setMerkleRoot`](https://github.com/mellow-finance/flexible-vaults/blob/c9c71818f0a8bead183f9860db3b8bbceb7b8f37/src/permissions/Verifier.sol#L140) authorizes callers through this vault role, so this is a direct control path outside the timelock proposer/executor flow. Please revoke the EOA's role or explicitly document and approve this trust-boundary exception.

2. **The audit annotation overstates the scope of the linked report.**

   Both the [vault metadata](https://github.com/lidofinance/diffyscan/blob/50113b1a2851ad34488ee3710383f3dfafd2fe8e/config_samples/robinhood/mainnet/meta/meta_usde_robinhood_config.yaml#L43-L46) and [core metadata](https://github.com/lidofinance/diffyscan/blob/50113b1a2851ad34488ee3710383f3dfafd2fe8e/config_samples/robinhood/mainnet/meta/meta_core_robinhood_config.yaml#L82-L85) label the March 2026 report as the “Latest audit” next to commit `c9c71818…`. That report covers only `SyncDepositQueue.sol` and `SyncQueue.sol` (134 LoC), while these configs also rely on vault, subvault, verifier, oracle, manager, and submitter code.

   Comparing the configured core commit with the prior core-audit commit leaves eight changed effective files, including `Vault.sol`, `Subvault.sol`, `Oracle.sol`, and `ShareModule.sol`. Some other components have their own scoped reports, but the core delta is not covered by the cited report. Please make the annotation accurately describe its limited scope and provide the missing audit/delta justification before treating the deployment as covered.

3. **The `V3Template` wildcard suppresses all bytecode-provenance checking.**

   [`"any": true`](https://github.com/lidofinance/diffyscan/blob/50113b1a2851ad34488ee3710383f3dfafd2fe8e/config_samples/ethereum/mainnet/vaults/vaults_voting_config.json#L80-L89) accepts every bytecode difference for this contract, and the [regression test](https://github.com/lidofinance/diffyscan/blob/50113b1a2851ad34488ee3710383f3dfafd2fe8e/tests/test_no_wildcard_regression.py#L28-L33) only exempts this entry from the no-wildcard rule.

   Re-running the base config with the wildcard removed produced exact source matches for all 3 contracts and exact bytecode matches for 2 of 3; `V3Template` remained uncovered after constructor simulation reverted, and Diffyscan exited non-zero. As written, future or unexpected bytecode drift for this contract will pass. Please replace the wildcard with pinned historical state or another granular, reproducible proof.

### Additional required coverage

4. **The new Robinhood configs are not exercised by CI.**

   The [regression workflow matrix](https://github.com/lidofinance/diffyscan/blob/50113b1a2851ad34488ee3710383f3dfafd2fe8e/.github/workflows/regression.yml#L12-L55) does not include either new config and only wires Ethereum/Hoodi RPC variables. The unit suite therefore does not validate the live Robinhood configurations. Please add both configs and the required Robinhood RPC environment.

tamtamchik:
1. Expected, see https://github.com/lidofinance/state-mate/pull/172 
2. Looks like you have checked not final commit for prev audit, diff with the final commit shows no differences in contracts (added explanation to the files)
3. V3Template is deprecated check and looks like we need to cleanup regressions (separate issue)
4. Robinhood is unstable and will give flaky tests, so for now it's better to keep it out of CI

TheDZhon:
Claude:

---

Two things remain:

1. A **live deployment defect** that this PR would certify as clean — an unrevoked deployer
   privilege. Tracked by the author in state-mate PR 172 but **not resolved**.
2. Two **PR-hygiene** items: unrelated Ethereum changes bundled in, one of which weakens an
   existing governance-critical check.

## Reconciliation of the three reviews

| # | Claim | Raised by | Author response | **My verdict after verification** |
|---|---|---|---|---|
| 1 | Extra EOA can replace verifier merkle root | this review + Codex | "Expected, see state-mate PR 172" | **CONFIRMED, open.** Response understates it — the author's own state-mate config marks it *must renounce*, not accepted. See F1. |
| 2 | Audit annotation overstates report scope | this review + Codex | "you checked not final commit; diff shows no differences" | **RESOLVED / Codex claim REFUTED.** Author is correct; Codex compared against the audit's *initial* commit. Verified below. |
| 3 | `V3Template` wildcard suppresses all bytecode checking | this review + Codex | "deprecated check, cleanup in separate issue" | **Open, disposition reasonable but incomplete.** If deprecated, delete the entry rather than wildcard it. See F2. |
| 4 | New configs not exercised by CI | Codex | "Robinhood unstable, would give flaky tests" | **Reasoning factually grounded** — I independently measured the instability. But it is the *explorer*, not the RPC, and the real fix is retry/backoff. See F5. |
| 5 | `constructor_calldata` never cross-checked against explorer | this review only | not addressed | **Open.** Tool gap; I closed it manually — all 16 entries byte-identical. See F3. |
| 6 | Dead `dependencies.contracts` entry | this review only | not addressed | **Open, still present at head** (lines 59-63). See F4. |

---

## F1 — CRITICAL: unrevoked deployer privilege on the live vault

**Status: open. Blocks earnUSDe accepting deposits. Not a defect in this PR — a defect this PR
would certify as clean.**

The deployer EOA `0xE98Be1E5538FCbD716C506052eB1Fd5d6fC495A3` still holds
`permissions.Verifier.SET_MERKLE_ROOT_ROLE` on the live Robinhood vault. It is an EOA (0 bytes of
code) and is **not in the actor inventory** supplied by the protocol team.

Evidence, reproduced against `rpc.mainnet.chain.robinhood.com` (chain 4663):

```
Verifier.SET_MERKLE_ROOT_ROLE()                   = 0xfc199f685d023b44b528c5fcb9cebfe292e64340dd5729b20761da4ad1e93024
keccak("permissions.Verifier.SET_MERKLE_ROOT_ROLE") = 0xfc199f685d023b44b528c5fcb9cebfe292e64340dd5729b20761da4ad1e93024   (identical)

vault.hasRole(role, 0xE98Be1E5...)                = true
vault.hasRole(role, timelock 0xa9BcE5B8...)       = true
vault.getRoleMemberCount(role)                    = 2

eth_call setMerkleRoot(0x..dead) from 0xE98Be1E5. = 0x                 (SUCCEEDS)
eth_call same from control 0x1111..1111           = revert 0xee90c468  (Forbidden)
```

`Verifier.setMerkleRoot` is gated purely on `vault().hasRole(SET_MERKLE_ROOT_ROLE, msg.sender)`
(`src/permissions/Verifier.sol:22-27,140` @ `c9c71818`), so this EOA can unilaterally replace the
merkle root that defines which calls the Curator may make from Subvault 0 — bypassing the intended
LazyAdmin -> Timelock path entirely.

### Why this is an oversight, not a design choice

1. The deployment calldata granted this EOA **five** roles: `DEFAULT_ADMIN`, `CREATE_SUBVAULT`,
   `SET_MERKLE_ROOT`, `ALLOW_SUBVAULT_ASSETS`, `SET_SUBVAULT_LIMIT`. **Four were revoked; this one
   was missed.** It has no residual privilege anywhere else — timelock
   proposer/canceller/executor/admin all `false`, vault `DEFAULT_ADMIN` `false`, FeeManager
   ownership moved to LazyAdmin.
2. **Plasma, Mantle and Ethereum earnUSD all revoked it.** Same role/EOA pair queried on each:
   all `false`. Robinhood is the only deployment where it persists.
3. **The author's own state-mate config says it must go.** state-mate PR 172,
   `configs/lido-earn/mainnet/earnusd-vaults-robinhood.yaml:382-383`:
   ```yaml
   # WARN: mellow_deployer must renounce SET_MERKLE_ROOT_ROLE (on-chain count is 2)
   *SET_MERKLE_ROOT_ROLE : [*earnUSDe_timelockController]
   ```
   The asserted member set is the timelock **alone**. The WARN was present in the **first** commit
   of that PR (`91397a6`, 2026-08-27T12:47Z) — before either review was posted, so the author found
   it independently. But "Expected" in the PR reply is the wrong disposition: state-mate encodes the
   *desired* state and flags a required renunciation, so that config is expected to **fail** against
   current chain state until the role is renounced.

### Exploitability and current exposure

Not solo-exploitable: execution still requires `CALLER_ROLE` (Curator 3/7), since `verifyCall`
rejects a `who` without it. So this is a **collusion and grief vector**, plus a governance-model
break — not a unilateral drain path.

Current exposure is **zero**: `merkleRoot = 0x0`, `allowedCalls = 0`, `queueLimit = 0`,
share `totalSupply = 0`. Nothing to steal today.

### Required action

Renounce/revoke `SET_MERKLE_ROOT_ROLE` from `0xE98Be1E5...` before queues open, then re-run
state-mate PR 172 to green. Until then, neither PR should be read as attesting the role model.

---

## F2 — MEDIUM: scope creep that weakens a governance-critical check

Two Ethereum configs unrelated to Robinhood are bundled into this PR:
`ethereum/mainnet/srv3/srv3_upgrade_template_config.yaml` and
`ethereum/mainnet/vaults/vaults_voting_config.json`.

The `"any": true` rule on `V3Template` (`0x34E01ecFebd403370b0879C628f8A5319dDb8507`) disables
**all** bytecode verification for a contract that is part of a live V3 vote.

Verified: the stated reason is accurate. Running that contract alone produced
`Failed to simulate deployment: execution reverted`, the rule matched, and **source comparison
still passed 75/75 files with 0 diffs**. So residual risk is confined to constructor/immutable
content, not source provenance.

Also verified: `diffyscan/utils/node_handler.py:35,72` hard-codes `"latest"` for both
`eth_getCode` and `eth_call`, and no `block` option exists in the config schema or CLI. So pinning
historical state — the fix both this review and Codex suggested — is **not currently possible**.

The author states the check is deprecated and cleanup belongs to a separate issue. Reasonable, but
the wildcard ships meanwhile. **A deprecated check should be deleted, not wildcarded**: a wildcard
silently passes future drift and reads as "verified" in the summary, whereas removing the contract
makes reduced coverage explicit. Recommend splitting both Ethereum files into their own PR so the
Robinhood config is not gated on this discussion.

The srv3 half of the change is *good* practice by contrast — granular `immutables` with specific
offsets and values, both verified against on-chain bytecode (see verified table).

---

## F3 — MEDIUM: the green run does not attest the init payloads

`diffyscan/utils/calldata.py:36-40`: when `constructor_calldata` is present in the config,
diffyscan uses it and **never compares it against the explorer's constructor arguments** — even
though it fetched them (`Explorer constructor calldata: 1568 bytes`).

For a `TransparentUpgradeableProxy`, only the `_admin` immutable reaches runtime bytecode. Every
substantive value in the `initialize(bytes)` payload — role grants, oracle security params,
timelock proposers/executors, fee recipient — is therefore **outside what the bytecode comparison
proves**. A wrong or malicious init payload in the config would produce a green run.

I closed the gap by hand: **all 16 declared entries are byte-identical to the explorer-reported
constructor arguments**, and every decoded value matches live state. So the config is materially
correct — but the tick-mark did not establish that.

Recommend an upstream equality assertion when both sources are available, and noting in the PR
that init payloads were verified out-of-band.

---

## F4 — LOW: dead dependency entry (still present at head)

`meta_usde_robinhood_config.yaml:59-63` declares a `dependencies.contracts` -> openzeppelin-contracts
mapping that is **never used** — the completed run resolved files only from `src` and `lib` origins
(252 `lib` + 116 `src`, zero `contracts`). It is also inconsistent with `meta_core_config.yaml` in
the same PR, which omits it. Carried over from the Plasma config.

---

## F5 — INFO: explorer instability is real; CI exclusion is grounded but the fix is retry

The author declined CI inclusion citing Robinhood instability. **Independently confirmed, with a
correction: it is the explorer, not the RPC.**

- `robinhoodchain.blockscout.com` returned HTTP 500 on roughly **1 in 5** requests across repeated
  probes of the same addresses; a one-shot v1 probe even reported a verified contract as
  unverified, then succeeded on retry.
- `diffyscan/utils/common.py:330` `fetch()` has **no retry**, so a single 5xx aborts the whole run.
  Eight consecutive attempts died this way, always at the first not-yet-cached address.
- The RPC endpoint itself was reliable throughout.

I could only obtain a complete run by temporarily patching a retry loop into `fetch()` in a scratch
clone (reverted afterwards; **not** part of the PR). Retry/backoff on explorer 5xx would make both
CI inclusion and manual runs viable, and is worth its own upstream issue. Meanwhile the PR should
carry a pasted successful run log, since reviewers cannot easily reproduce one.

---

## F6 — INFO: items for conscious sign-off (not defects)

- `TimelockController.getMinDelay() = 0` — no delay despite the name. Matches Plasma.
- RiskManager vault limit = `2^254-1` — effectively unbounded. Matches Plasma.
- `queueLimit = 0` and 0 queues live -> deposits/redemptions are not yet possible. The config will
  need `DepositQueue`/`RedeemQueue` entries once created (cf. `ethereum/mainnet/meta/meta_usd_config.yaml`,
  which lists them).
- Live `ACCEPT_REPORT_ROLE` sits on the OracleSubmitter *contract*, not on LazyAdmin directly —
  but LazyAdmin is the accepter inside that contract, so the indirection matches the actor
  inventory rather than contradicting it.
- YAML list indent in the srv3 addition (`- reason` at 4 spaces) differs from the sibling entry
  (6 spaces). Cosmetic; both parse.

TheDZhon (approved):

PR #206, opened 2026-08-27
feat: add diffyscan configs for EDF on Hoodi

chasingrainbows:
## Summary

Add diffyscan configs to verify the EDF/DSM v5 deployment on Hoodi (vote #83, LIP-37).

## Changes

- `config_samples/ethereum/hoodi/edf/hoodi_edf_config.json` - DelegationFactory and 17 DelegationContracts from lidofinance/execution-delegation-framework. Constructor args are pinned in the config because the contracts were deployed through the factory, so the explorer has no constructor args for them. Values were read from chain and cross-checked with deployed-hoodi.json and the vote script.
- `config_samples/ethereum/hoodi/edf/hoodi_edf_core_config.json` - DepositSecurityModule v5 and the new LidoLocator implementation from lidofinance/core (feat/edf). DSM needs `deployment_from` (its constructor reverts when the sender is the zero address) and an `allowed_diffs` rule for six immutable positions: the ATTEST/PAUSE/UNVET message prefix hashes depend on address(this), so they cannot match in the deployment simulation. The pinned values equal the on-chain getters.
- `config_samples/ethereum/hoodi/edf/hoodi_edf_easy_track_config.json` - SetDepositsReserveTarget factory from lidofinance/easy-track (deposit-reserve branch). Allowed diffs: four import lines rewritten by brownie verification, plus the CBOR metadata hash; the executable bytecode is identical. Same pattern as the existing vaults and srv3 easy track configs.
- `.gitignore` - add node_modules.

The new configs use the Etherscan V2 API (`api.etherscan.io` + `explorer_chain_id: 560048`), since the V1 endpoint used by the older Hoodi configs is deprecated.

## Testing

- All three configs pass a full diffyscan run (source and bytecode): EDF config - 18/18 exact matches; core config - LidoLocator exact match, DSM matches with the documented immutable diff; easy track config - passes with the two granular allowed diffs.
- `uv run pytest -q` - 374 passed.

## Related Issues

Linear: STC-889


F4ever (approved):

TheDZhon (commented):
GG, CI failures are unrelated to the scope of this PR.

TheDZhon (approved):
📦

Ziars (approved):

PR #207, opened 2026-08-27
docs: rework README and add reference guides

tamtamchik:
## Summary

- align the README layout with state-mate
- add a project banner and concise quick start
- move usage, configuration, CLI, bytecode, and allowlist guidance into docs

## Verification

- checked local Markdown links
- parsed all YAML examples
- loaded the quick-start config with Diffyscan
- ran commit hooks


TheDZhon:
Codex:

Found three documentation issues:

- **[P2] `fail_on_bytecode_comparison_error` behavior is misstated** — [`docs/configuration.md:38`](https://github.com/lidofinance/diffyscan/blob/28007903aecc09c3a0adca637acee2fbe450d325/docs/configuration.md#L38)

  Setting this to `false` does not make common compilation, calldata, or RPC errors non-fatal. Those exceptions are always recorded as `status="failed"` and cause exit code 1. The flag currently affects the outer exception handler instead. Either fix the implementation or document the actual behavior.

- **[P2] Not all linked sample configurations are working** — [`docs/configuration.md:5`](https://github.com/lidofinance/diffyscan/blob/28007903aecc09c3a0adca637acee2fbe450d325/docs/configuration.md#L5)

  Several Soneium and Unichain samples provide only `explorer_hostname_env_var`, but runtime code never resolves that field and requires `explorer_hostname`. These samples fail even when the named environment variable is set. Avoid calling the entire directory “working,” or implement hostname resolution.

- **[P3] Libraries are incorrectly grouped with per-contract overrides** — [`docs/bytecode-comparison.md:23`](https://github.com/lidofinance/diffyscan/blob/28007903aecc09c3a0adca637acee2fbe450d325/docs/bytecode-comparison.md#L23)

  `constructor_args`, calldata, caller, and extra sources are keyed per contract; `libraries` is a global source-path mapping applied to every contract. Clarify this, particularly because different deployments can use different addresses for the same library.

TheDZhon:
Claude:

**Verdict: approve with changes.** This is a clear net improvement — it documents 9 CLI flags and 4 config fields that had no coverage before, and everything it asserts about current behaviour holds up. The issues are almost all things the rewrite dropped, plus one claim that is false for the install path the docs recommend.

## Findings

**1. `.env` guidance is false for the install path the docs recommend** — medium

`diffyscan.py:826` calls bare `load_dotenv()`. python-dotenv resolves its search root from the *caller module's* `__file__`, not the CWD, so for an installed CLI it walks up from `site-packages` and never reaches the user's working directory.

Reproduced — `uv tool install .`, then in a fresh directory containing a valid `.env`:

```
ValueError: Explorer token not found in env ("ETHERSCAN_EXPLORER_TOKEN")
```

Isolated: `find_dotenv()` → `''`, `find_dotenv(usecwd=True)` → the CWD path.

`docs/configuration.md` states this unconditionally ("Diffyscan loads a local `.env` file before reading credentials"). `docs/how-to.md` scopes it to "Run a local checkout", where it *is* true — but "Install the CLI" gives no env guidance at all, and this PR removed the old README's `export VAR=…` snippet, which is the only form that works for an installed CLI. So someone following "Install the CLI" has no documented way to supply credentials.

Fix either side: restore the `export` form under "Install the CLI", or change the code to `load_dotenv(find_dotenv(usecwd=True))` and keep the docs as written.

**2. Dev Container setup dropped entirely** — medium

`.devcontainer/` still exists, and `regression.yml` builds that Dockerfile and runs *all* CI inside it. The old README called it the recommended path; neither the new README nor `docs/how-to.md` mentions it now, so the one environment with CI parity is undiscoverable.

**3. `pre-commit install` dropped** — medium

`.pre-commit-config.yaml` enforces conventional commit messages through a commit-msg hook, plus black and mypy on pre-commit. The new README says to *run* the hooks (`uv run pre-commit run --all-files`) but never to install them, so nothing enforces them locally.

Since `default_install_hook_types` is set, this is one line and no longer needs the old `--hook-type` flags:

```sh
uv run pre-commit install
```

**4. `any: true` exclusivity is undocumented** — low

`allowed_diffs.py:571-577` raises `cannot combine any with other allowlist facets`. Both facet tables in `docs/bytecode-comparison.md` list `any: true` as a peer of `immutables` / `cbor_metadata` / `byte_ranges`, so a reader would reasonably combine them and hit a config error. Worth a note in the tables or just below them.

**5. Linked-library troubleshooting note dropped** — low

The old README's hint — if simulation fails with `Invalid params`, or Diffyscan reports `unlinked libraries`, a library address is missing or keyed by the wrong file. Both strings are still live in the code (`explorer.py:778`, `node_handler.py:10`), and this is hard to rediscover from the error alone. `docs/bytecode-comparison.md` keeps the correct keying rule but not the symptom that leads you to it.

**6. README "What it checks" blurs the trust boundary** — low

> - linked libraries, constructor calldata, compiler settings, and EVM version

These are inputs Diffyscan *takes from the explorer* to reproduce the build, not things it verifies. `docs/bytecode-comparison.md` gets this right ("Diffyscan reuses explorer-provided constructor calldata, linked-library addresses, and EVM version"). For a verification tool the trusted-input list is the interesting part, so it's worth wording the README bullet the same way.

**7. Wildcard-test scope** — low

> The regression test … rejects new wildcard rules unless the repository records an explicit justification.

The test only globs `config_samples/**`, so it says nothing about a user's own config. Same imprecision as the old README, but the new phrasing reads more like a general guarantee.

## Nits

- `--no-support-brownie` is a no-op: `--support-brownie` uses `BooleanOptionalAction` with no `default=`, so the default is already `None`. Listing it in the table gives it equal weight with a flag that does something.
- Cache-clearing commands (`rm -rf .diffyscan_cache/`) dropped; useful and cheap to keep in `docs/how-to.md`.
- `docs/configuration.md` says JSON is accepted but shows no JSON example, though most files in `config_samples/` are JSON.
- `pyproject.toml` has `readme = "README.md"`, so the relative links and the banner break in the packaged long description. Low impact while distribution is git-only.

## Pre-existing, outside this diff

Not introduced here, but two of them touch lines this PR wrote:

- **`explorer_hostname_env_var` is dead.** Declared at `custom_types.py:76` and set in 9 sample configs, but never read anywhere; `explorer_hostname` is asserted required at `diffyscan.py:633`. Verified with `L1_EXPLORER_API_HOSTNAME` exported:

  ```
  [WARN] Failed to find "explorer_hostname" in the config
  AssertionError: explorer_hostname is required
  ```

  Affected: `unichain/mainnet/{L1,L2,L2_gov}`, `soneium/mainnet/{L1,L2,L2_gov}`, `soneium/testnet/{L1,L2,L2_gov}`. `docs/configuration.md` is right to omit the dead field, but "Working configurations are under `config_samples/`" isn't quite true until the configs or the code are fixed.
- `network`, `audit_url`, and `metadata` are likewise declared in `Config` but never read.
- `diffyscan.py:850` tells users to "specify a path with `--path`", but the argument is positional. `docs/cli.md` now documents it correctly, which makes the runtime string the outlier — a one-line code fix would align them.

tamtamchik:
Thanks for the detailed reviews. I’ve addressed the remaining feedback:

- the actual `fail_on_bytecode_comparison_error` behavior documented;
- the scope of library overrides, `any: true`, wildcard checks, and explorer-provided inputs clarified;
- the Dev Container, pre-commit, cache cleanup, and linked-library troubleshooting guidance restored;
- credential guidance for installed and local usage added;
- `explorer_hostname_env_var` resolution and added test coverage implemented;
- README links and the incorrect `--path` message fixed;
- `network`, `audit_url`, and `metadata` documented as informational fields.

TheDZhon:
Codex: approve, no remaining findings :clap:

TheDZhon:
Claude: approve, nitpicks only (won't bother with them in this PR)

TheDZhon (approved):

PR #216, opened 2026-09-07
docs: refresh verification skills and add allowlist workflow

tamtamchik:
Repository skills contained stale config requirements, references to removed verifier functions, and guidance that could hide unverified bytecode. Update the four existing workflows against the current implementation and add a dedicated `allowed-diffs` workflow for justified exceptions and wildcard tightening.

Keep detailed explorer payload mappings and diagnostic recipes in linked references, including an executable immutable-rule coverage check. Preserve manual invocation settings and argument hints, and update the workflow index in `CLAUDE.md`.

Clarify the limits of source hunk rules, the prerequisites for constructor overrides inside allowlist rules, and the deployment-simulation exception accepted by bytecode wildcards. Keep config review separate from edits and live diagnostics, and correct the architecture description of immutable analysis. Explain recursive filename lookup, missing-source suggestions, token fallback and the simulation gas setting.

## Validation
- Five skills: metadata and local reference links validated.
- Immutable-rule example executed: expected difference accepted and five unrelated changes rejected.
- 69 config, allowlist, explorer and HTTP tests passed.
- Independent reviews checked runtime accuracy, workflow completeness and removed guidance; reproduced that source rules match coordinates without pinning text.
- A further read-only Claude audit checked all five skills against runtime code and tests; confirmed findings were corrected after inspecting the implementation.
- Six read-only scenarios exercised through Claude CLI; follow-up runs checked CREATE2 reasoning, metadata evidence, hostname matching, config review and report fields. These checks do not establish live-chain support or reliability across models.

### DIFF
--- a/.claude/skills/add-explorer/SKILL.md
+++ b/.claude/skills/add-explorer/SKILL.md
@@ -1,155 +1,47 @@
 ---
 name: add-explorer
-description: Add support for a new blockchain explorer API to diffyscan. Use when a new chain or explorer type needs to be supported.
+description: Adds or repairs Diffyscan explorer API routing and response adapters for a new host, chain or payload format. Use when explorer support is missing or dispatches incorrectly. If an existing adapter supports the deployment, continue with new-config instead of changing code.
+argument-hint: "[explorer-name]"
 disable-model-invocation: true
-argument-hint: [explorer-name]
 ---
 
-Guide through adding support for a new blockchain explorer to diffyscan.
+Find the smallest change that supports the requested explorer. Run commands from the repository root.
 
-## Decision tree: do you need code changes?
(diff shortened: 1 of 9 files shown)


TheDZhon (approved):
No actionable findings in [PR #216](https://github.com/lidofinance/diffyscan/pull/216) (https://github.com/lidofinance/diffyscan/pull/216), reviewed at 2d01b85. Rechecked GitHub; the head is unchanged.

The revised workflows match the runtime behavior, including allowlist restrictions, constructor overrides, and simulation-error handling.

Validation:

- 410 tests passed
- Five skill metadata records and 32 local links/anchors validated
- Embedded immutable-rule example passed all six assertions

